PR #25, opened 2015-02-23
Shorthand font style doesn't parse

mprudhom:
The font shorthand "12em monospace" doesn't parse. The test case should have picked this up, but the validation API call was wrong. The attached PR fixes the test case validation so it now fails, but it doesn't fix the source of the shorthand parsing error.


mprudhom:
I fixed the error. It was a typo (".len" instead of ".length"). :cry:

chad3814:
:cry: thanks

chad3814:
0.2.23 was published with the fix

PR #31, opened 2015-06-02
Fixing bug w/font shorthand returns "background"

arvind:


arvind:
There seems to be an additional error somewhere, because I'm getting `font: sans-serif 11px 11px;` as the output, rather than `font: 11px sans-serif;`

chad3814:
probably a conflict between font-size accepting a length and line-height also accepting a length, so they are both taking it; which isn't right (I think). maybe line-height shouldn't be part of the font shorthand, I'll have to look at some docs

PR #35, opened 2016-01-13
Improve handling of padding and margin 

dpvc:
This pull request improves the handling of `padding` and `margin`.  They now set and clear the `-top`, `-right`, `-bottom`, and `-left` properties properly, and validate the value they are set to.  It also adds some tests to verify that this works.

As part of the update, the `implicitSetter()` function is simplified, and made more efficient (the `parser` function is only called once on any of the parts).  Also, unneeded empty parts are not retained in the `_values` array.  Finally, `getParts()` now allows multiple spaces as separators.


chad3814:
Awesome, I'll review this in the morning CST and then merge and push a new version to npm

Issue #37, opened 2016-01-21
padding/margin not completely fixed


domenic:
https://github.com/chad3814/CSSStyleDeclaration/pull/35 helped but it seems like the sub-properties are not properly passed to the parent.

Test case from jsdom (with latest csstyle). Mostly passing except the one failing line.

```js
const document = jsdom.jsdom();

for (const prop of ["padding", "margin"]) {
  document.body.style[prop] = "1px 2px 3px 4px";

  t.strictEqual(document.body.style[prop], "1px 2px 3px 4px");
  t.strictEqual(document.body.style[prop + "Top"], "1px");
  t.strictEqual(document.body.style[prop + "Right"], "2px");
  t.strictEqual(document.body.style[prop + "Bottom"], "3px");
  t.strictEqual(document.body.style[prop + "Left"], "4px");

  document.body.style[prop + "Top"] = "1em";
  document.body.style[prop + "Right"] = "2em";
  document.body.style[prop + "Bottom"] = "3em";
  document.body.style[prop + "Left"] = "4em";
  t.strictEqual(document.body.style[prop], "1em 2em 3em 4em"); // fails, left as 1px 2px 3px 4px

  document.body.style[prop] = "1mm";
  t.strictEqual(document.body.style[prop], "1mm");
  t.strictEqual(document.body.style[prop + "Top"], "1mm");
  t.strictEqual(document.body.style[prop + "Right"], "1mm");
  t.strictEqual(document.body.style[prop + "Bottom"], "1mm");
  t.strictEqual(document.body.style[prop + "Left"], "1mm");

  document.body.style[prop] = "1% 2%";
  t.strictEqual(document.body.style[prop], "1% 2%");
  t.strictEqual(document.body.style[prop + "Top"], "1%");
  t.strictEqual(document.body.style[prop + "Right"], "2%");
  t.strictEqual(document.body.style[prop + "Bottom"], "1%");
  t.strictEqual(document.body.style[prop + "Left"], "2%");

  document.body.style[prop] = "3pc 2pc 1pc";
  t.strictEqual(document.body.style[prop], "3pc 2pc 1pc");
  t.strictEqual(document.body.style[prop + "Top"], "3pc");
  t.strictEqual(document.body.style[prop + "Right"], "2pc");
  t.strictEqual(document.body.style[prop + "Bottom"], "1pc");
  t.strictEqual(document.body.style[prop + "Left"], "2pc");
}

t.done();
```

[...]

PR #38, opened 2016-01-27
Patch padding margin

dpvc:
This pull request adds login to set `padding` and `margin` when one of its sub-properties changes (and all four sub-properties are set).  This should resolve issue #37.  It also adds a new test to check that it does what it is supposed to.  I also fixed some tab-versus-spacing.

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -623,21 +623,21 @@ exports.subImplicitSetter = function (prefix, part, isValid, parser) {
     return function (v) {
         if (typeof v === 'number') v = v.toString();
         if (typeof v !== 'string') return undefined;
-	if (!isValid(v)) return undefined;
-	v = parser(v);
-	this._setProperty(property,v);
-	var parts = [];
+        if (!isValid(v)) return undefined;
+        v = parser(v);
+        this._setProperty(property,v);
+        var parts = [];
         for (var i = 0; i < 4; i++) {
-	    if (this._values[subparts[i]] == null || this._values[subparts[i]] === '') break;
-	    parts.push(this._values[subparts[i]]);
+            if (this._values[subparts[i]] == null || this._values[subparts[i]] === '') break;
+            parts.push(this._values[subparts[i]]);
         }
-	if (parts.length === 4) {
-	    for (i = 0; i < 4; i++) {
-		this.removeProperty(subparts[i]);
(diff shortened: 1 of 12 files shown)


chad3814:
Does this only work if you specify all four properties to a `margin:` or `padding:`? I'll pull and mess around today

dpvc:
If all four of `padding-top`, `padding-right`, `padding-bottom`, and `padding-left` are set when one of them is specified, then the `padding` style that combines them should be set to the combined value (and the individual values should not show in the `cssText` string, just the combined one).  It does require all four individual values to be set for that to happen.

For example, of you did

```
style.padding = "1px 2px 3px 4px";
style.paddingTop = "10px";
```

then `style.padding` should show `10px 2px 3px 4px` and `style.cssText` should be `padding: 10px 2px 3px 4px` not something like `padding: 1px 2px 3px 4px; padding-top: 10px`.  But if you did

```
style.paddingTop = "10px";
style.paddingBottom = 20px";
```

and the `paddingLeft` and `paddingRight` are not specified elsewhere, then `padding` will be blank and `style.cssText` should be `padding-top: 10px; padding-bottom: 20px`.

Does that answer your question?

chad3814:
Looks like I missed in your earlier PR but you can't set `padding` or `margin` to `0` anymore. That's a simple fix in their `isValid()`, I'll commit that and make a new test for it.

One thing is, we need to cascade `removeProperty()` to take out the subproperties.
```
style.padding = 0;
style.paddingLeft = '1px';
style.cssText; // 'padding: 0px 0px 0px 1px;'
style.removeProperty('padding');
style.cssText; // ''
style.paddingLeft; // '1px'
style.paddingTop; // '0px'
```
So need to remove all the sub-properties. I think I'll merge this and file an issue for sub-property removal.

What do you think about taking over this project @dpvc ?

dpvc:
> One thing is, we need to cascade `removeProperty()` to take out the subproperties.

Good point.  I'm afraid I don't know what everything is supposed to do, so didn't think of that.

> What do you think about taking over this project?

Sorry. I've got too much on my plate as it is.  Wish I could do more.

chad3814:
no worries, 0.2.32 is published with this fix

dpvc:
Great!  Thanks!

Issue #40, opened 2016-04-22
width and height value of auto is not supported


dpvc:
Setting a style of `auto` for `width` or `height` doesn't work.  This is because the use of `parseMeasurement()` in the `set` method doesn't pass `auto`, so nothing is done.

Here is an example:

```
var cssstyle = require("./lib/CSSStyleDeclaration");
var style = new cssstyle.CSSStyleDeclaration();
style.width = "auto";
console.log(style);
console.log(style.width);
console.log(style.cssText);
```

This prints three blank lines rather than the expected two lines of `auto` followed by `width: auto;`.

chad3814:
published in cssstyle@0.2.36; sorry I missed the issue and PR

PR #41, opened 2016-04-22
Allow width and height to be set to auto

dpvc:
Allow `width` and `height` to be set to `auto`, and add new test to verify the fix.  Resolves issue #40.


domenic:
ping @chad3814

PR #43, opened 2016-08-13
Use babel to inline code for properties

ForbesLindesay:
This inlines the code for css properties, renaming variables as necessary to avoid conflicts.

This reduces the time taken to `require('CSSStyleDeclaration')` between 80ms and 150ms in my tests.  I attribute most of this speedup to the fact that it no longer needs to run the expensive module resolution algorithm for each of the properties.


cpojer:
This is awesome. To give some context – we use this through jsdom inside of Jest and the require time of this module alone is about 100ms which is about 20 % of jsdom's overall require time. We'd like to speed up Jest's startup time because it is imperative for a unit test framework to run quickly. Please consider merging this :)

chad3814:
so, that looks great, I'm just concerned that if something goes wrong with it later, I am not smart enough to fix it

cpojer:
Don't worry, we are always here to help you :)

chad3814:
published version 0.2.37 with the change. if either of you want to take it over, let me ;)

PR #47, opened 2016-12-14
Add support for numeric values passed to ‘border-spacing’, e.g. ‘border-spacing: 0’.

jansiegel:
When providing `0` to `border-spacing`, cssstyle throws a `v.toLowerCase is not a function` error.

### DIFF
--- a/lib/properties/borderSpacing.js
+++ b/lib/properties/borderSpacing.js
@@ -10,6 +10,7 @@ var parse = function parse(v) {
     if (v === '' || v === null) {
         return undefined;
     }
+    v = v + '';
     if (v.toLowerCase() === 'inherit') {
         return v;
     }


lsiu:
We use jsdom for some testing and some libraries we test with has border-spacing set to 0. 
Can we merge this PR and publish a new release to npmjs?
Thanks!

lsiu:
I did another PR #52 which include a test and follow the same approach used in [padding.js](https://github.com/chad3814/CSSStyleDeclaration/blob/master/lib/properties/padding.js)

sairoutine:
+1 for merging this, got the same error on jsdom.

@chad3814

jansiegel:
#52 does basically the same thing as this one + adds tests, so I'm +1 for merging either one of those.

PR #49, opened 2017-02-15
Add transform property

piec:
Hi,

I use jsdom to pre-render a dom with d3 but it's missing the transform styles
This commit adds support for transform

Also mentioned in tmpvar/jsdom#1321 and NV/CSSOM#78

Cheers

### DIFF
--- a/lib/properties/transform.js
+++ b/lib/properties/transform.js
@@ -1,0 +1,12 @@
+'use strict';
+
+module.exports.definition = {
+    set: function (v) {
+        this._setProperty('transform', v);
+    },
+    get: function () {
+        return this.getPropertyValue('transform');
+    },
+    enumerable: true,
+    configurable: true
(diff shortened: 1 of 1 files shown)


hraban:
Is this planned for merging ever? I'm currently using 

```js
mynode.attributes['style'].value += ' transform: translate(-50%, -50%);';
```

which is a bit of an eye sore :P

Hikariii:
What is the problem with merging this?

Hikariii:
A temporary work-around can be this:
```js
Object.defineProperty(jsDom.window._core.CSSStyleDeclaration.prototype, 'transform', {
	get : function () {
		return this.getPropertyValue('transform');
	},
	set : function (v) {
		this._setProperty('transform', v);
	},
	enumerable: true,
	configurable: true
});
```

PR #50, opened 2017-07-07
Add test for background property

andy128k:
Test fails but it shows that background property doesn't work for positions.
It seems shorthandParser doesn't fit well for background property.


jsakas:
Closing stale PR. If this is still needed, please reopen a PR.

PR #52, opened 2017-09-27
An alternative fix for PR#47 - Add support for numeric values passed to ‘border-spacing’, e.g. ‘border-spacing: 0’

lsiu:
This fix include a test and follow the approach used in [padding.js](https://github.com/chad3814/CSSStyleDeclaration/blob/master/lib/properties/padding.js)

### DIFF
--- a/lib/properties/borderSpacing.js
+++ b/lib/properties/borderSpacing.js
@@ -10,6 +10,9 @@ var parse = function parse(v) {
     if (v === '' || v === null) {
         return undefined;
     }
+    if (typeof v === 'number') {
+        v = String(v)
+    }
     if (v.toLowerCase() === 'inherit') {
         return v;
     }
--- a/lib/properties/borderSpacing.js
+++ b/lib/properties/borderSpacing.js
@@ -10,8 +10,8 @@ var parse = function parse(v) {
     if (v === '' || v === null) {
         return undefined;
     }
-    if (typeof v === 'number') {
-        v = String(v)
+    if (v === 0) {
+        return "0px";
     }
     if (v.toLowerCase() === 'inherit') {
         return v;
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ CSSStyleDeclaration is a work-a-like to the CSSStyleDeclaration class in Nikita 
 
 Why not just issue a pull request?
 ----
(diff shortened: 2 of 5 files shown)


eddies:
@lsiu can you rebase this on master? Hopefully that would increase the chances of getting this merged. I have a rebased version of your branch ready at https://github.com/eddies/CSSStyleDeclaration/tree/border-spacing-numeric-values, but I don't want to duplicate your PR

lsiu:
If you refer to the [w3c border-spacing spec](https://www.w3.org/wiki/CSS/Properties/border-spacing),  you can see the "initial value" is 0, not "0", so I suppose a numerical value is also a valid, and should be taken into consideration.

lib/properties/borderSpacing.js
```diff
@@ -10,6 +10,9 @@ var parse = function parse(v) {
     if (v === '' || v === null) {
         return undefined;
     }
+    if (typeof v === 'number') {
+        v = String(v)
```
jsakas: I think `return undefined` makes more sense to mimic the current browser behaviour.

jsakas:
@lsiu @eddies @jansiegel

Browsers do not seem to support this functionality. Tested in Chrome, FF, Safari, Edge:

```javascript
var table = document.createElement('table')
console.log(table.style.borderSpacing); // ""
table.style.setProperty('border-spacing', 10);
console.log(table.style.borderSpacing); // ""
table.style.setProperty('border-spacing', '10');
console.log(table.style.borderSpacing); // ""
table.style.setProperty('border-spacing', '10px');
console.log(table.style.borderSpacing); // 10px
```

Instead of converting the value, I think it would make more sense to return if the value is not a string.

eddies:
@jsakas In Firefox 61.0 (OS X):
```
var table = document.createElement('table')
console.log(table.style.borderSpacing); // ""
table.style.setProperty('border-spacing', 10);
console.log(table.style.borderSpacing); // "10px"
table.style.setProperty('border-spacing', '10');
console.log(table.style.borderSpacing); // "10px"
```

Moreover, in Firefox, Chrome (67) and Safari (11.1.1), I observe the following:
```
var table = document.createElement('table')
console.log(table.style.borderSpacing); // ""
table.style.setProperty('border-spacing', 0);
console.log(table.style.borderSpacing); // "0px"
table.style.setProperty('border-spacing', '0');
console.log(table.style.borderSpacing); // "0px"
```

So I don't think a blanket `return undefined` is the clear winner. I think it's either the PR as-is or a special case for `0` and `undefined` for other numbers.

jsakas:
@eddies interesting, it seems to work for `0` only, but not for any other integer.

PR #53, opened 2017-12-06
Ensure ex length units are correctly parsed

bpartridge:
Per https://www.quirksmode.org/css/units-values/ the 'ex' unit has widespread cross-browser support, and it's even relied on by Gmail in styling its "quoted text" bars to send within emails. However, it was missing from the parser, causing JSDOM to silently drop any style rule that used ex values - not the easiest thing to debug! We've added a quick fix to the regex, as well as unit tests that fail without the changed regex and succeed with it.

Hopefully this can be a quick merge and minor release!


jsakas:
We have a similar PR open for rem... https://github.com/chad3814/CSSStyleDeclaration/pull/55

Doesn't look like this is being maintained anymore.

PR #56, opened 2018-05-04
added flex-related properties and a test

ignacionr:
This will allow dependent projects such as jsdom to keep flex-related values.


jsakas:
Thank you @ignacionr

PR #62, opened 2018-05-22
cross-platform `npm test` script

reyronald:
The previous `npm test` script executed a `.sh` file
that isn't natively supported by Windows.

Since the actual code inside the `.sh` is pretty small,
it can be completely embedded inside a `npm` script instead,
making it cross-platform and more straight-forward.

_This is a suggestion, if there are specific reasons why this should be in a `.sh` file as it was, disregard this PR._

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "resolve": "~1.1.7"
   },
   "scripts": {
-    "test": "./scripts/run_tests.sh",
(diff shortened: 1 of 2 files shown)


jsakas:
@reyronald thanks for the PR. No reason specifically that I am aware of except that the original author was developing on *nix.

PR #69, opened 2018-08-15
Support vh and vw for lengths

dallashall-procore:
Adds the ability to specify lengths using viewport measurements.

Note: does not add the ability to use `vmax`.

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -56,9 +56,9 @@ module.exports = {
     'Test From Style String': function (test) {
         var style = new cssstyle.CSSStyleDeclaration();
         test.expect(8);
-        style.cssText = 'color: blue; background-color: red; width: 78%';
-        test.ok(3 === style.length, 'length is not 3');
-        test.ok('color: blue; background-color: red; width: 78%;' === style.cssText, 'cssText is wrong');
+        style.cssText = 'color: blue; background-color: red; width: 78%; height: 50vh;';
+        test.ok(4 === style.length, 'length is not 4');
+        test.ok('color: blue; background-color: red; width: 78%; height: 50vh;' === style.cssText, 'cssText is wrong');
         test.ok('blue' === style.getPropertyValue('color'), "getPropertyValue('color') failed");
         test.ok('color' === style.item(0), 'item(0) failed');
         test.ok('background-color' === style[1], 'style[1] failed');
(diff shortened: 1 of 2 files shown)


tests/tests.js
```diff
@@ -108,13 +108,17 @@ module.exports = {
...
         style.height = 6;
         test.ok('' === style.height, 'height does not remain unset');
         style.width = 0;
         test.ok('0px' === style.width, 'width is not 0px');
         style.height = '34%';
         test.ok('34%' === style.height, 'height is not 34%');
+        style.height = '100vh';
+        test.ok('100vh' === style.height, 'height is not 100vh');
+        style.height = '100vw';
+        test.ok('100vw' === style.height, 'height is not 100vw');
```
jsakas: These assertion names are misleading. We are testing that the height **is** 100vh / 100vw. So it should read `height is 100vw` or `height: 100vw` ... can you update the two new tests and the two above it, please?

tests/tests.js
```diff
@@ -56,9 +56,9 @@ module.exports = {
     'Test From Style String': function (test) {
         var style = new cssstyle.CSSStyleDeclaration();
         test.expect(8);
-        style.cssText = 'color: blue; background-color: red; width: 78%';
-        test.ok(3 === style.length, 'length is not 3');
-        test.ok('color: blue; background-color: red; width: 78%;' === style.cssText, 'cssText is wrong');
+        style.cssText = 'color: blue; background-color: red; width: 78%; height: 50vh;';
+        test.ok(4 === style.length, 'length is not 4');
+        test.ok('color: blue; background-color: red; width: 78%; height: 50vh;' === style.cssText, 'cssText is wrong');
```
jsakas: I don't know what the reasoning behind the original code was here but the assertion names seem wrong here too. 

This assertion name is incorrect "length is not 3"...

```
test.ok(3 === style.length, 'length is not 3');
```

It should be 

```
test.ok(3 === style.length, 'length is 3');
```
🤔

jsakas: Ah, yes that would actually make sense as to why its worded as a failure message. The docs for nodeunit don't really specify so it might have just been the preference of the original author and maintainers. 

In that case, lets just leave it with the style of the other tests until we can swap out the test runner to something that is not deprecated.

dallashall-procore: If you have a particular testing library you like, I'd be willing to help make that switch.

jsakas: In other projects I use jest or mocha/chai. Would like to use either of these.

I have an issue for this here https://github.com/jsakas/CSSStyleDeclaration/issues/65

PR #71, opened 2018-09-02
Added new script to download CSS properties from W3C standard.

Havunen:
Hey!

I found this plugin is missing `float` property and started to investigate how to add it. I noticed there is pre-defined hardcoded list of CSS2 (+ some CSS3) properties added to this library. I thought it would be cool to use W3C json property list, instead of hard coded list in the test. 

Doc:
https://www.w3.org/Style/CSS/all-properties.en.html

I created a simple script that downloads the list of CSS properties and filters out the properties which are currently in `Working Draft` state and/or later. 

This list of properties is then used as basis for the tests.
Downloading new list can be done by running command `npm run download`

After downloading and generating the property list, tests can be run. The current state of tests is 
`FAILURES: 1068/1909 assertions failed (398ms)` about 50% success.

Fixing these tests would take this module to support all W3C recommended styles. It kinda resolves this too https://github.com/jsakas/CSSStyleDeclaration/issues/32

...
I just needed float and it little bit escalated... :smile:


jsakas:
@fatfisz it seems tests are now failing on master after this merge.

fatfisz:
Yes, I'm fixing that right now. It's easier to merge this to master and fix it there quick (I had to sleep in the meantime though) than engage in a back-and-forth that could take more time. Expect the fix in 1 hour!

fatfisz:
Fixed.

PR #72, opened 2018-09-23
Added more Flex-related styles.

step-off:
Hello there! 
First of all, thank you for awesome work!

By this PR I'm adding some more flex styles:
flex-grow, flex-shrink, flex-basis and flex shorthand.

There is nothing special in this changes, but I'd like to clearify one moment. 
When I added flex shorthand and write tests on it, none of them was 'green'. It turns out, that setting flex shorthand value, for example, `0 1 250px` would give output `1 1 250px`. Flex-grow is being reassigned by flex-shrink value, because it is absolutly valid for it and their type and possible values are identical. 
To resolve this confusion I decided to pass shorhand part's index as second parameter in subprop `isValid` function while runnig shorthandParser. Now it works fine. Tests prove it.

Nonetheless, if you'll see this solution as not clear enough, I'll gladly fix or rewrite this according to your advice :)

Hope it helps somehow. 
Best regards

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,0 +1,4 @@
+module.exports.POSITION_AT_SHORTHAND = {
+    first: 0,
+    second: 1
+}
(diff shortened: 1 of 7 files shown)


jsakas:
Thanks for the PR!

Moved to https://github.com/jsakas/CSSStyleDeclaration/pull/80 so I could resolve conflicts.

PR #74, opened 2019-02-02
Adds CSS Color Module 4 colors and hsl(a)

### LINKED ISSUE #48: Colors are not interpreted correctly
**Keywords**
CSS has hundreds of color keywords, but CSSStyle only recognizes a few
https://github.com/chad3814/CSSStyleDeclaration/blob/master/lib/parsers.js#L100-L116

**HSL and HSLA**
CSSStyle has no logic to interpret HSL or HSLA values. Here's the regex for Hex, RGA, and RGBA. HSL and HSLA are missing.

https://github.com/chad3814/CSSStyleDeclaration/blob/master/lib/parsers.js#L28-L30

**Computing the result**
Hex values are correctly converted to RGB, however keywords (even valid ones) are not. If CSSStyle intends to emulate a browser's behavior, it should convert everything to RGB or RGBA

**TLDR**
CSSStyle doesn't handle color values correctly

bennypowers:
fixes #48


fatfisz:
Great work! Would you be willing to modify it a bit - because having a big switch statement like that is unwieldy - by adding a script similar to the ones existing in `scripts/`? The output would be a JSON file with an array containing all the scraped colors (the source being https://www.w3.org/TR/css-color-4/ if I'm not mistaken).

bennypowers:
Well i already scraped the colours (and I don't think colors module 4 will change) , so do you want me to just create that file in `lib/named_colors.json`?

bennypowers:
Ok @fatfisz see if latest changes suit your liking

fatfisz:
Ok, let's go with that. I'll need to adapt the parser anyway because of `var()`, but it will be better done as a separate change, because `rgb` and `rgba` will be affected too.

bennypowers:
Awesome. Please advise when released, since they're waiting on this in https://github.com/GoogleChrome/lighthouse/issues/7139

fatfisz:
Sure. I just became a contributor to add some stuff that I need at work (and help with some tidying up), so I'm driven to get the new release out quickly :) Expect it under a week or so.

bennypowers:
👋 Just a friendly bump. any chance we can cut a release?

fatfisz:
Hi, I did what I could, now we're waiting for the release (I don't have permissions). Thanks for your patience, this should be released soon!

jsakas:
@bennypowers @fatfisz v1.2.0 is published to npm.

Thanks for your help!

PR #82, opened 2019-02-16
fix: remove the extra comma that would cause error in some old apps

### LINKED ISSUE #81: Extra comma
The extra comma here caused this bug in my application: https://github.com/jsakas/CSSStyleDeclaration/blob/master/lib/CSSStyleDeclaration.js#L201

```
/app/node_modules/cssstyle/lib/CSSStyleDeclaration.js:202
       );
       ^

SyntaxError: Unexpected token )
   at createScript (vm.js:56:10)
   at Object.runInThisContext (vm.js:97:10)
```

It works normally as long as I removed the extra comma. Would be great if there could be a quick patch for the issue. Thanks!

loking:
To fix #81

### DIFF
--- a/lib/CSSStyleDeclaration.js
+++ b/lib/CSSStyleDeclaration.js
@@ -198,7 +198,7 @@ Object.defineProperties(CSSStyleDeclaration.prototype, {
         this.setProperty(
           dummyRule[i],
           dummyRule.getPropertyValue(name),
-          dummyRule.getPropertyPriority(name),
+          dummyRule.getPropertyPriority(name)
         );
       }
       this._onChange(this.cssText);


fatfisz:
Hi, there's a problem with that branch: we have some warnings in the build because after running prettier again the comma would be restored. Could you first add `--max-warnings 0` after "eslint" in the `lint` and `lint:fix` commands?

After that you could change `trailingComma` in `.eslintrc.js` to "es5" and run `npm run lint:fix` - this will catch more trailing commas.

fatfisz:
Orrr wait, I'll do the first part, then just refresh the branch and set "es5" as `trailingComma`.

fatfisz:
Ok, you can now merge the current master into your branch.

PR #88, opened 2019-04-13
Add support to hex alpha transparency (#rgba)

GabrielBibiano:
This pattern should not be much used, but I believe the support in the project is ideal.
https://developer.mozilla.org/en-US/docs/Web/CSS/color#Syntax

Forgive any gaffe in the code or with English. This is my first contribution to an open source project and I am also not fluent in the language.


lib/parsers.js
```diff
@@ -26,7 +26,7 @@ var lengthRegEx = /^(0|[-+]?[0-9]*\.?[0-9]+(in|cm|em|mm|pt|pc|px|ex|rem|vh|vw))$
 var percentRegEx = /^[-+]?[0-9]*\.?[0-9]+%$/;
 var urlRegEx = /^url\(\s*([^)]*)\s*\)$/;
 var stringRegEx = /^("[^"]*"|'[^']*')$/;
-var colorRegEx1 = /^#[0-9a-fA-F][0-9a-fA-F][0-9a-fA-F]([0-9a-fA-F][0-9a-fA-F][0-9a-fA-F])?$/;
+var colorRegEx1 = /^#[0-9a-fA-F][0-9a-fA-F][0-9a-fA-F]([0-9a-fA-F])?([0-9a-fA-F][0-9a-fA-F][0-9a-fA-F])?([0-9a-fA-F][0-9a-fA-F])?$/;
```
fatfisz: The expression got a lot less readable, though it wasn't at a good state even before.

I suggest something like so instead: `/^#([0-9a-fA-F]{3,4}){1,2}$/`.

lib/parsers.js
```diff
@@ -287,15 +287,30 @@ exports.parseColor = function parseColor(val) {
     alpha = 1;
   var parts;
   var res = colorRegEx1.exec(val);
-  // is it #aaa or #ababab
+  // is it #aaa, #ababab, #aaaa, #abababaa
   if (res) {
+    var defaultHex = val.substr(1);
     var hex = val.substr(1);
-    if (hex.length === 3) {
+    if ([3, 4].indexOf(defaultHex.length) > -1) {
```
fatfisz: If there are only two values to compare, having the comparison in the code explicitly instead of the much more complicated expression should work better.

I suggest using `hex.length === 3 || hex.length === 4` instead, it reads better.

lib/parsers.js
```diff
@@ -287,15 +287,30 @@ exports.parseColor = function parseColor(val) {
...
       hex = hex[0] + hex[0] + hex[1] + hex[1] + hex[2] + hex[2];
+
+      if (defaultHex.length === 4) {
```
fatfisz: While it won't impact performance too negatively, having an additional copy of a string just to check its length is a bit confusing. The logic will be a bit more clear if we just save the length at the beginning:

```
var hex = val.substr(1);
var length = hex.length;
```

and then we can use that in the previous (3 and 4 comparison) and this (4 comparison) check.

lib/parsers.js
```diff
@@ -287,15 +287,30 @@ exports.parseColor = function parseColor(val) {
...
+        hex = hex + defaultHex[3] + defaultHex[3];
+      }
     }
     red = parseInt(hex.substr(0, 2), 16);
     green = parseInt(hex.substr(2, 2), 16);
     blue = parseInt(hex.substr(4, 2), 16);
+    if (hex.length === 8) {
+      var hexAlpha = hex.substr(6, 2);
+      var hexAlphaToRgbaAlpha = (parseInt(hexAlpha, 16) / 255).toFixed(2);
```
fatfisz: Just wrapping this in one more `parseInt` would be abit more straightforward than adding a condition below:

```
var hexAlphaToRgbaAlpha = parseInt(parseInt(hexAlpha, 16) / 255).toFixed(2));
```

or even something like this:

```
var hexAlphaToRgbaAlphaUnrounded = parseInt(hexAlpha, 16) / 255);
var hexAlphaToRgbaAlpha = parseInt(hexAlphaToRgbaAlphaUnrounded.toFixed(2));
```

lib/parsers.js
GabrielBibiano: In fact I also use it to capture the default alpha and duplicate it to the end result because the 'hex' variable is already modified. What to do in this case?

`
if (defaultHex.length === 4) {
    hex = hex + defaultHex[3] + defaultHex[3];
}
`

lib/parsers.js
GabrielBibiano: Much better. Thanks

lib/parsers.js
GabrielBibiano: Sure.

lib/parsers.js
GabrielBibiano: But in doing so, if the alpha is between 0 and 1, like 0.67, it will be rounded and should not. I make this encoding below just in case it's 0 or 1, so I'm going to remove the decimal places.

GabrielBibiano:
@fatfisz Something about the considerations?

lib/parsers.js
jsakas: @GabrielBibiano based on the w3 spec and testing in Chrome / FF, I think it should be rounded to the nearest 1000th. So use `toFixed(3)`.

Using the `Number` constructor seems to remove unwanted decimal places (e.g. converts 0.500 to 0.5) which I think is what is intended by the second conditional being discussed. So we can use something like:

```
var hexAlphaToRgbaAlpha = Number((parseInt(hexAlpha, 16) / 255).toFixed(3))`
```

GabrielBibiano: Right! I'll make this modification.

jsakas:
@fatfisz ping ping

Issue #100, opened 2019-08-28
Doesn't support `calc`


malash:
```js
import cssstyle from "cssstyle";

export const stringifyCssText = (styleObject: object): string => {
  let cssText = "";
  const style = new cssstyle.CSSStyleDeclaration(newCssText => {
    cssText = newCssText;
  });
  Object.assign(style, styleObject);

  return cssText;
};
console.log(
  stringifyCssText({
    width: "100vh",
    height: "calc(100vh * 2)"
  })
);
// will output `width: 100vh;` but `height` is missing
```

Demo here https://codesandbox.io/s/flamboyant-moore-983md

malash:
cc @Jimexist @byunicorn

jsakas:
Hi @malash, thanks for reporting. I think the lib should definitely support calc properties. I will take a look.

jsakas:
Digging in, it looks like supporting calc isn't exactly trivial. Browsers will actually evaluate what it can when setting the result. 

I put up a WIP branch that handles expressions that cannot be evaluated directly, such as percentage types or units that do not match. But in your example, the result of the calc would actually just be `calc(200vh)`. So we need to parse the calc function and evaluate the expression if possible - this is where it starts to get a little more complicated than a simple patch on the lib.

Some example tests:

```
style.setProperty('width', 'calc(100px * 2)');
expect(style.getPropertyValue('width')).toEqual('calc(200px)');
```

```
style.setProperty('width', 'calc(calc(100px * 2) - calc(100vw + 80px))');
expect(style.getPropertyValue('width')).toEqual('calc(200px - (100vw + 80px))');
```

https://github.com/jsdom/cssstyle/pull/101

aeagle:
Hi @jsakas. I can see that you've merged this but haven't created an NPM release yet that includes it which means dependant issues in jsdom remain outstanding, i.e. https://github.com/jsdom/jsdom/issues/1332. 

A test against the latest version `2.2.0` confirms that support for calc is still not there - https://runkit.com/embed/2f7g0k4iu8nx.

Were you intending to release only after adding logic to evaluate the calc expression?

jsakas:
Hi @aeagle - v2.3.0 has been published this morning with the basic calc implementation.

aeagle:
Fantastic. Many thanks!

PR #101, opened 2019-09-01
WIP: Implement calc

jsakas:
Begin to implement the `calc()` css function.

This work is not complete, because calc will evaluate what it can inside of the expression and return that as the result. e.g. `100% * 2` becomes `200%`. There is a failing test case on the branch for this.

We need to parse the expression, and units, and evaluate them according to the spec and existing browser implementations.

https://drafts.csswg.org/css-values-3/#calc-notation
https://developer.mozilla.org/en-US/docs/Web/CSS/calc

### DIFF
--- a/lib/CSSStyleDeclaration.test.js
+++ b/lib/CSSStyleDeclaration.test.js
@@ -547,4 +547,10 @@ describe('CSSStyleDeclaration', () => {
     expect(style.getPropertyValue('--foo')).toEqual('');
     expect(style.getPropertyValue('--fOo')).toEqual('purple');
   });
+
+  test('supports calc', () => {
+    const style = new CSSStyleDeclaration();
+    style.setProperty('width', 'calc(100% - 100px)');
+    expect(style.getPropertyValue('width')).toEqual('calc(100% - 100px)');
+  });
 });
--- a/lib/CSSStyleDeclaration.test.js
+++ b/lib/CSSStyleDeclaration.test.js
@@ -552,5 +552,8 @@ describe('CSSStyleDeclaration', () => {
     const style = new CSSStyleDeclaration();
     style.setProperty('width', 'calc(100% - 100px)');
     expect(style.getPropertyValue('width')).toEqual('calc(100% - 100px)');
+
+    style.setProperty('width', 'calc(100px * 2)');
+    expect(style.getPropertyValue('width')).toEqual('calc(200px)');
   });
 });
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -18,6 +18,7 @@ exports.TYPES = {
   ANGLE: 8,
   KEYWORD: 9,
   NULL_OR_EMPTY_STR: 10,
+  CALC: 11,
 };
 
 // rough regular expressions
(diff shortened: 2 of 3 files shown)


nitinthewiz:
This is a really interesting PR. I'm using jsdom to create elements in node and then save the HTML. I don't know if this is the right tool for the job, but it does it quite well, specially with .serialize(), which is just awesome! 

I needed calc, because it is needed in the output, and the browser doesn't need calc to be computed, as you need it to be.

So, I'm going to adopt this code for my needs.

I just have two questions for you - 
1. Would you recommend some other library for creating DOM elements?

2. Would you recommend a way for me to patch this code on top of jsdom? I was just going to fork the latest, put this patch on, and always point my package.json to my own repo.

jsakas:
@nitinthewiz that is an interesting use case for this library, thanks for sharing. I have been debating whether or not I think that calc should be computed. It seems like a lot of work and I'm not actually sure if that feature would be beneficial. 

I think if we implemented calc in a way that it could be set but not computed, it would solve your use case as well as close https://github.com/jsdom/cssstyle/issues/100

I may have been overthinking it when I was exploring computing the values in a way that browsers do.

nitinthewiz:
@domenic heh, *technically* browsers compute calc, but I don't see a good reason for jsdom to, because it doesn't have a view...

snoozbuster:
I just spent upwards of a couple hours trying to debug why adding a very simple `width: calc(30% - 2px)` to one of my vuejs components caused the style attribute to disappear entirely in my tests (which are run under nodejs with jest and jsdom), causing them to fail... only to trace it back to this issue/PR. My use case is also one which would be solved by a simple set-but-not-compute solution; right now I have to test that my code is computing the width correctly but can't test that it actually binds to the element (a crucial part of this code's functionality). Is there a possibility that this PR can be revived and merged?

jsakas:
Closing in favor of https://github.com/jsdom/cssstyle/pull/115

If we need to add computed calc in the future it would be a larger initiative, but for now I think we should at least let calc values pass through.

cdoublev:
I'm in a situation where all I need is having the calc values passed through, but I would like to set a `border` length with it, and [`parserLength`](https://github.com/jsdom/cssstyle/blob/b527ed722364dc6d156487c652df100572075dee/lib/properties/borderWidth.js#L10) [does not handle it currently](https://github.com/jsdom/cssstyle/blob/b527ed722364dc6d156487c652df100572075dee/lib/parsers.js#L185) (only the "measurement" parser handles calc, ie. values which are either length or percentages). `parseInteger`, `parseNumber`, `parsePercent`, `parseColor`, and `parseAngle` should also handle it.

While this is a valid issue that may be relatively easy to fix, I took a look at computing calc before.

My biggest concern for doing it is that some properties will have a percentage value resolved at computed value time, while some others will have them resolved at used value time, which may involve resolving other properties, eventually from a containing element. There's also resolving `attr()` nested in calc.

So my question is: is it worth trying a partial implementation to compute calc, ie. returning as is a calc that includes a percentage, but that conforms to the other requirements, ie. with [types checking](https://drafts.csswg.org/css-values-3/#calc-type-checking)? Or should I only create an issue / pull a fix request for the cases mentionned above where calc should also pass through?

PR #107, opened 2019-10-25
Add support for `ch` unit

tinovyatkin:
Adds support for standard `ch` unit: https://caniuse.com/#feat=ch-unit


jsakas:
Thanks for the PR!

PR #108, opened 2019-11-01
Update script that generates allProperties.js

jsakas:
Update the output syntax to match the most recent updates


scripts/download_latest_properties.js
```diff
@@ -68,13 +68,14 @@ request(url, function(error, response, body) {
     );
     out_file.write('/*\n *\n * https://www.w3.org/Style/CSS/all-properties.en.html\n */\n\n');
 
-    out_file.write('var allProperties = new Set();\n');
```
sirian: Maybe it's better to replace with `out_file.write('module.exports = new Set(' + JSON.stringify(CSSpropertyNames.map(camelToDashed), null, 2) + ');'`

Issue #89, opened 2019-05-23
CSS Custom Properties not supported


jcousins-ynap:
CSS Custom properties are ignored as are the rules which use them.

```js
const
  cssstyle = require('cssstyle'),
  customProp = new cssstyle.CSSStyleDeclaration(),
  normalProp = new cssstyle.CSSStyleDeclaration();

customProp.cssText = `--hello-world: blue; background: var(--hello-world);`
console.log(customProp.background) // "" -- ❌ expected "blue"

normalProp.cssText = `background: blue;`
console.log(normalProp.background) // "blue" -- ✅
```

Unlike the CSS color names PR, I have no idea how to resolve this issue.
I don't mind contributing to solve this issue, but I will likely need some help.

ckedwards:
If the point is to match `CSSStyleDeclaration` then this kind of resolution is not necessary.

For example, when I tried this in chrome:
```js
var styleObj = document.styleSheets[0].cssRules[0].style;
styleObj.cssText = "--hello-world: blue; background: var(--hello-world);"
console.log(styleObj.background)
// Output: var(--hello-world)
```

It seems like custom properties just need to be set on the `CSSStyleDeclaration`  even though they aren't in allProperties or allExtraProperties.

A custom property is defined as "any valid identifier that starts with two dashes, except -- itself"
https://drafts.csswg.org/css-variables-1/#custom-property

sheijne:
Honestly, why has this issue been open for so long? There are also multiple issues describing this problem in the jsdom repo, which go back multiple years..

I've been looking through the cssstyle source and it appears that this issue should be easy to solve.

Correct me if I'm wrong, but the problem seems to be with the way properties are filtered:

https://github.com/jsdom/cssstyle/blob/master/lib/CSSStyleDeclaration.js#L60
```
if (!allProperties.has(lowercaseName) && !allExtraProperties.has(lowercaseName)) {
```

According to the spec:

> A custom property is any property whose name starts with two dashes (U+002D HYPHEN-MINUS), like --foo. The <custom-property-name> production corresponds to this: it’s defined as any valid identifier that starts with two dashes. Custom properties are solely for use by authors and users; CSS will never give them a meaning beyond what is presented here.
- https://www.w3.org/TR/css-variables/#custom-property

So a solution could be:

```
if (lowercaseName.startsWith('--') || (!allProperties.has(lowercaseName) && !allExtraProperties.has(lowercaseName))) {
```


[...]

PR #111, opened 2020-01-23
feat: Implement read/write of custom properties

eps1lon:
Part of #89 which is blocking https://github.com/facebook/react/pull/17896.

I did not implement full support since this requires more work and is only partially useful for JSDOM because resolution of values of the css variables requires the cascade. 

Browser behavior: https://codesandbox.io/s/css-custom-properties-access-423ni


jsakas:
@eps1lon thanks for the PR! I think a lot of folks will be happy about this one. 

I will merge and bump to version v2.2.0

PR #112, opened 2020-01-23
fix: webkit properties not being recognized

eps1lon:
Seems like they were never tested. I'm not sure if the two added tests can be fixed with independent commits.

### DIFF
--- a/scripts/generate_implemented_properties.js
+++ b/scripts/generate_implemented_properties.js
@@ -9,8 +9,15 @@ const camelToDashed = require('../lib/parsers').camelToDashed;
 const dashedProperties = fs
   .readdirSync(path.resolve(__dirname, '../lib/properties'))
   .filter(propertyFile => propertyFile.substr(-3) === '.js')
-  .map(propertyFile => camelToDashed(propertyFile.replace('.js', '')));
-
+  .map(propertyFile => camelToDashed(propertyFile.replace('.js', '')))
+  .map(property => {
+    const isVendorSpecific = /^(o|moz|ms|webkit)-/.test(property);
+    if (isVendorSpecific) {
+      return '-' + property;
+    } else {
+      return property;
+    }
+  });
 const out_file = fs.createWriteStream(path.resolve(__dirname, '../lib/implementedProperties.js'), {
   encoding: 'utf-8',
 });
--- a/lib/CSSStyleDeclaration.test.js
+++ b/lib/CSSStyleDeclaration.test.js
@@ -568,5 +568,5 @@ describe('CSSStyleDeclaration', () => {
     // TODO: this should be a number
     expect(style.WebkitLineClamp).toEqual('20');
     expect(style.getPropertyValue('-webkit-line-clamp')).toEqual('20');
-  })
+  });
 });
--- a/lib/CSSStyleDeclaration.test.js
+++ b/lib/CSSStyleDeclaration.test.js
@@ -547,4 +547,20 @@ describe('CSSStyleDeclaration', () => {
     expect(style.getPropertyValue('--foo')).toEqual('');
     expect(style.getPropertyValue('--fOo')).toEqual('purple');
   });
+
+  test('implements webkitTextFillColor', () => {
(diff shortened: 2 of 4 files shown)


jsakas:
@eps1lon is this a regression we missed in the last PR that ended up being 2.2.0 - or new functionality?

scripts/generate_implemented_properties.js
```diff
@@ -9,8 +9,15 @@ const camelToDashed = require('../lib/parsers').camelToDashed;
 const dashedProperties = fs
   .readdirSync(path.resolve(__dirname, '../lib/properties'))
   .filter(propertyFile => propertyFile.substr(-3) === '.js')
-  .map(propertyFile => camelToDashed(propertyFile.replace('.js', '')));
-
+  .map(propertyFile => camelToDashed(propertyFile.replace('.js', '')))
+  .map(property => {
+    const isVendorSpecific = /^(o|moz|ms|webkit)-/.test(property);
```
eps1lon: Sneaking in support for other vendor prefixes. I'm not sure if you're opinionated about other vendor prefixes. I'm looking at this purely from a testing perspective where I'd like to see support for all vendor prefixes. Otherwise it's hard to write tests that cover all browsers at once.

eps1lon:
> @eps1lon is this a regression we missed in the last PR that ended up being 2.2.0 - or new functionality?

Tried the tests in 2.1.0 (`git checkout v2.1.0`) and they were failing as well. I couldn't find any tests related to webkit properties. I can try to create some tests related to webkit properties that are passing in 2.1.0.

domenic:
Only webkit properties are in the spec; other engine prefixes should not be added.

eps1lon:
> Only webkit properties are in the spec; other engine prefixes should not be added.

This statement clashes with 

https://github.com/jsdom/cssstyle/blob/ebd2dab41a3396419513a7698cdd37494727d0da/lib/allExtraProperties.js#L4-L5

eps1lon:
@jsakas Is there anything I can do to get this merged?

jsakas:
@eps1lon sorry for the delay here. Are you able to resolve the conflict for me? Also, based on @domenic's comment I think only webkit should be supported at this time.

eps1lon:
> Are you able to resolve the conflict for me?

Sure thing.

> Also, based on @domenic's comment I think only webkit should be supported at this time.

Personally I would prefer if this also handled other prefixed properties. This is in line with 
https://github.com/jsdom/cssstyle/blob/ebd2dab41a3396419513a7698cdd37494727d0da/lib/allExtraProperties.js#L4-L5

But I understand if you want to change that stance to "we only support properties in the spec".

scripts/generate_implemented_properties.js
ExE-Boss: The real bug is in: <https://github.com/jsdom/cssstyle/blob/62b3b1bfbeb6cdebf829de9caf777635e68c605d/lib/parsers.js#L712>

It should be:
```js
var first_segment = /^([^-]+)-/;
```

nyroDev:
Hi @eps1lon, do you intend to fix conflict for this pull request?
I encounter an error with a webkit CSS proprty not being recognized correctly, and this issue should fix it.
I you're not willing to do it, I'll be happy to do it.

What I also noticed is that webkit specific value are accessible through 2 names:  
webkitTextFillColor and WebkitTextFillColor

I'm currenlty not sure if this PR will implement both of them (test case looks like it is)

nyroDev:
BTW allProperties define [only 1 webkit property](https://github.com/jsdom/cssstyle/blob/master/lib/allProperties.js#L11) alone, `-webkit-line-clamp`
I think this PR should remove this line, [introduced lately on January](https://github.com/jsdom/cssstyle/commit/054d45605531f6d4c1c4fab43efeff9142ec0f72#diff-2b14c78d944e247cf5f913665d604f2f4b8b090ba39c76b66b9b25140a065f1e)

eps1lon:
Not working on this one at the moment

PR #116, opened 2020-04-01
Migrate `CSSStyleDeclaration` to WebIDL2JS

ExE-Boss:
This migrates `CSSStyleDeclaration` to use **WebIDL2JS**.

Part of <https://github.com/jsdom/jsdom/issues/2727>.

---

There’s also <https://github.com/jsdom/webidl2js/issues/188>, which would simplify the generation of `CSSStyleDeclaration‑properties.webidl` by allowing the resulting string to be passed to **WebIDL2JS** directly, instead of having to write it to a temporary file.


scripts/convert-idl.js
```diff
@@ -0,0 +1,89 @@
...
+const propertyNames = Array.from(allProperties)
+  .concat(
+    Array.from(allExtraProperties).filter(prop => prop !== 'css-float' && !allProperties.has(prop))
+  )
+  .sort();
+
+const camelCasedAttributes = propertyNames.map(n => parsers.dashedToCamelCase(n));
+const webkitCasedAttributes = propertyNames
+  .filter(n => n.startsWith('-webkit-'))
+  .map(n => parsers.dashedToCamelCase(n.substring(1)));
```
TimothyGu, line 24: If you implement the full [CSS property to IDL attribute](https://drafts.csswg.org/cssom/#css-property-to-idl-attribute) algorithm as I commented above, the `.substring(1)` could be replaced by setting the _lowercase first_ flag.

scripts/convert-idl.js
```diff
@@ -0,0 +1,89 @@
...
+    path.resolve(__dirname, '../src/CSSStyleDeclaration-properties.webidl'),
+    {
+      encoding: 'utf-8',
+    }
+  );
+
+  const dateToday = new Date();
+  genIDL.write(
+    `// autogenerated - ${dateToday.getMonth() +
+      1}/${dateToday.getDate()}/${dateToday.getFullYear()}
```
TimothyGu, line 40: Let's make it scarier.

```suggestion
  genIDL.write(
    `// autogenerated by scripts/convert-idl.js. do not edit! ${new Date().toISOString()}
```

src/CSSStyleDeclaration.webidl
```diff
@@ -0,0 +1,16 @@
...
+  [CEReactions] attribute CSSOMString cssText;
+  readonly attribute unsigned long length;
+  getter CSSOMString item(unsigned long index);
+  CSSOMString getPropertyValue(CSSOMString property);
+  CSSOMString getPropertyPriority(CSSOMString property);
+  [CEReactions] void setProperty(CSSOMString property, [TreatNullAs=EmptyString] CSSOMString value, optional [TreatNullAs=EmptyString] CSSOMString priority = "");
+  [CEReactions] CSSOMString removeProperty(CSSOMString property);
+  readonly attribute CSSRule? parentRule;
+  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString cssFloat;
+};
```
TimothyGu, line 16: Say 
```suggestion
};

// Additional partial interfaces are generated by `scripts/convert-idl.js`.
```

lib/CSSStyleDeclaration-impl.js
```diff
@@ -9,23 +9,24 @@ var allExtraProperties = require('./allExtraProperties');
...
+    this._globalObject = globalObject;
+    this._values = {};
+    this._importants = {};
+    this._length = 0;
+    this._onChange =
+      onChangeCallback ||
+      function() {
+        return;
+      };
+  }
```
TimothyGu, line 29: While at it, let's simplify this a bit.

```suggestion
  constructor(globalObject, args, { onChangeCallback = () => {} }) {
    this._globalObject = globalObject;
    this._values = {};
    this._importants = {};
    this._length = 0;
    this._onChange = onChangeCallback;
  }
```

lib/CSSStyleDeclaration-impl.js
```diff
@@ -9,23 +9,24 @@ var allExtraProperties = require('./allExtraProperties');
...
-    function() {
-      return;
-    };
-};
-CSSStyleDeclaration.prototype = {
-  constructor: CSSStyleDeclaration,
+class CSSStyleDeclaration {
+  /**
+   * @constructor
+   * @see http://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSSStyleDeclaration
```
TimothyGu, line 17: Old! Looks like the rest of the file all have this though, so let's defer this to another PR.

lib/CSSStyleDeclaration.test.js
```diff
@@ -9,6 +9,7 @@ var parsers = require('./parsers');
 
 var dashedProperties = [...allProperties, ...allExtraProperties];
 var allowedProperties = dashedProperties.map(parsers.dashedToCamelCase);
+allowedProperties.push('cssFloat');
```
TimothyGu, line 12:
```suggestion
var allowedProperties = ['cssFloat', ...dashedProperties.map(parsers.dashedToCamelCase)];
```

ExE-Boss: The downside of your approach is that it creates an intermediate array in the `map` function, which is then immediately iterated over and discarded, thus having a `O(n) + O(n)` complexity, because it calls `map` and `[Symbol.iterator]`, whereas mine has `O(n) + O(1)` complexity, because it calls `map` and `push`.

---

The way your approach would remain `O(n)` would be to use:
```suggestion
var allowedProperties = ['cssFloat', ...dashedProperties.values().map(parsers.dashedToCamelCase)];
```

But that won't work until [**Iterator Helpers**](https://github.com/tc39/proposal-iterator-helpers) are implemented.

ExE-Boss:
Well, [the build passed](https://travis-ci.org/github/jsdom/cssstyle/builds/673674605), but it’s not reflected on **GitHub** due to `cssstyle` using the legacy service‑based **Travis‑CI.org** instead of the **GitHub App**‑based **Travis‑CI.com**.

TimothyGu:
travis-ci.com seems to have been down for maintenance.

scripts/convert-idl.js
```diff
@@ -0,0 +1,88 @@
...
+  for (const webkitCasedAttribute of webkitCasedAttributes) {
+    genIDL.write(`  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _${webkitCasedAttribute};
+`);
+  }
+
+  genIDL.write(`
+  // https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-dashed_attribute
+`);
+
+  for (const dashedAttribute of dashedAttributes) {
```
TimothyGu, line 64: I think it makes more sense do something like

```js
for (const property of propertyNames) {
  if (!property.includes('-')) continue;
  genIDL.write(…);
}
```

It fits better with what the spec says:

> For each CSS property _property_ that is a supported CSS property, except for properties that have no "`-`" (U+002D) in the property name, the following partial interface applies where _dashed attribute_ is _property_.

For WebKit above,

```js
for (const property of propertyNames) {
  if (!property.startsWith('-webkit-')) continue;
  const webkitCasedAttribute = cssPropertytoIDLAttribute(property, /* lowercaseFirst = */ true);
  genIDL.write(…);
}
```

scripts/generate_implemented_properties.js
```diff
@@ -4,12 +4,14 @@ const fs = require('fs');
...
 const t = require('babel-types');
 const generate = require('babel-generator').default;
-const camelToDashed = require('../lib/parsers').camelToDashed;
+const { idlAttributeToCSSProperty } = require('../lib/parsers');
 
 const dashedProperties = fs
   .readdirSync(path.resolve(__dirname, '../lib/properties'))
-  .filter(propertyFile => propertyFile.substr(-3) === '.js')
-  .map(propertyFile => camelToDashed(propertyFile.replace('.js', '')));
+  .filter(propertyFile => propertyFile.slice(-3) === '.js')
```
TimothyGu, line 11:
```suggestion
  .filter(propertyFile => propertyFile.endsWith('.js'))
```

lib/CSSStyleDeclaration-impl.js
```diff
@@ -48,7 +50,7 @@ CSSStyleDeclaration.prototype = {
    * @param {string} [priority=null] "important" or null
    * @see http://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSSStyleDeclaration-setProperty
    */
-  setProperty: function(name, value, priority) {
+  setProperty(name, value, priority) {
     if (value === undefined) {
```
TimothyGu, line 54: There are more things you can clean up here. For example, `value` can now only be a string.

TimothyGu:
Also an interesting question would be do we want to add some custom reflection support for CSS attributes? I think it might increase memory usage, but would allow us to get rid of the getter/setter methods on `CSSStyleDeclaration.prototype`. Would be interesting to experiment.

lib/CSSStyleDeclaration-impl.js
ExE-Boss: Actually, a good chunk of the implemented properties call `_setProperty` with `value = undefined`, expecting it to do nothing.

TimothyGu: Yeah, I understand. That's why I commented at this function in particular. I haven't checked other functions either fyi.

lib/CSSStyleDeclaration-impl.js
```diff
@@ -123,49 +128,32 @@ CSSStyleDeclaration.prototype = {
...
-};
+  }
+
+  [idlUtils.supportsPropertyIndex](index) {
+    return index >= 0 && index < this._length;
+  }
+}
 
-Object.defineProperties(CSSStyleDeclaration.prototype, {
+Object.defineProperties(CSSStyleDeclarationImpl.prototype, {
```
ExE-Boss, line 156: Changing this to be defined as getters in the class body pushed this PR over the edge of what git will consider a renamed file.

TimothyGu: Don't let tools dictate your code :) Worst case we could just turn on `-w` and `-M` to make Git try harder.

index.js
```diff
@@ -0,0 +1,37 @@
...
+webidlWrapper.install(sharedGlobalObject);
+
+const origCSSStyleDeclaration = sharedGlobalObject.CSSStyleDeclaration;
+
+/**
+ * @constructor
+ * @param {((cssText: string) => void) | null} [onChangeCallback]
+ * The callback that is invoked whenever a property changes.
+ */
+function CSSStyleDeclaration(onChangeCallback = null) {
```
TimothyGu, line 15: I was trying to say to use `undefined` instead of `null` as the sentinel for the absence of a callback.

```suggestion
function CSSStyleDeclaration(onChangeCallback = undefined) {
```

index.js
```diff
@@ -0,0 +1,37 @@
...
+  if (new.target === undefined) {
+    throw new TypeError('Illegal invocation');
```
TimothyGu, line 17: Let's give a more useful message.

```suggestion
    throw new TypeError('Class constructor CSSStyleDeclaration cannot be invoked without 'new'');
```

index.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (onChangeCallback !== null) {
+    onChangeCallback = webidlConversions.Function(onChangeCallback, {
+      context: 'Failed to constructor CSSStyleDeclaration: parameter 1',
+    });
+  }
```
TimothyGu, line 24: Please don't use Web IDL conversions if it's not actually a Web IDL method. If/when we implement IDL callback functions in webidl2js properly this function might disappear, as [`Function`](https://heycam.github.io/webidl/#Function) is a typedef in Web IDL.

README.md
```diff
@@ -4,12 +4,44 @@ A Node JS implementation of the CSS Object Model [CSSStyleDeclaration interface]
...
+This module default-exports the `CSSStyleDeclaration` interface constructor, with the change that it can be constructed and takes an `onChangeCallback` optional parameter:
+
+```ts
+export = class CSSStyleDeclaration {
+  /**
+   * @param onChangeCallback The callback that is invoked whenever a property changes.
+   */
+  constructor(onChangeCallback?: ((cssText: string) => void) | null);
+}
+```
```
TimothyGu, line 30: Can we stick to JavaScript? It's strange to have TypeScript in the README of a JavaScript project, and also I think more words to explain what this is is better than fewer.

```suggestion
This module default-exports the `CSSStyleDeclaration` interface constructor, with the change that it can be constructed with an optional `onChangeCallback` parameter. Whenever any CSS property is modified through an instance of this class, the callback (if provided) will be called with a string that represents all CSS properties of this element, serialized. This allows the embedding environment to properly reflect the style changes to an element's `style` attribute.

Here is a crude example of using the `onChangeCallback` to implement the `style` property of `HTMLElement`:
```js
const CSSStyleDeclaration = require('cssstyle');

class HTMLElement extends Element {
  constructor() {
    this._style = new CSSStyleDeclaration(newCSSText => {
      this.setAttributeNS(null, "style", newCSSText);
    });
  }
  get style() {
    return this._style;
  }
  set style(text) {
    this._style.cssText = text;
  }
}
<!-- the suggestion block doesn't let me type this: ``` -->
```

README.md
```diff
@@ -4,12 +4,44 @@ A Node JS implementation of the CSS Object Model [CSSStyleDeclaration interface]
...
+The `privateData` argument takes the following properties:
+
+```ts
+interface CSSStyleDeclarationPrivateData {
+  /**
+   * The callback that is invoked whenever a property changes.
+   */
+  onChangeCallback?: ((cssText: string) => void) | null
+}
+```
```
TimothyGu, line 47:
```suggestion
The `privateData` parameter of `create` and `createImpl` provides a way to specify the `onChangeCallback` that is a constructor parameter in the default export. Only the `onChangeCallback` property is supported on `privateData` currently, with the same semantics as the constructor parameter documented above.
```

README.md
```diff
@@ -4,12 +4,44 @@ A Node JS implementation of the CSS Object Model [CSSStyleDeclaration interface]
...
+  /**
+   * @param onChangeCallback The callback that is invoked whenever a property changes.
+   */
+  constructor(onChangeCallback?: ((cssText: string) => void) | null);
+}
+```
+
+### `cssstyle/webidl2js-wrapper` module
+
+This module exports the `CSSStyleDeclaration` [interface wrapper API](https://github.com/jsdom/webidl2js#for-interfaces) generated by [webidl2js](https://github.com/jsdom/webidl2js).
```
TimothyGu, line 34:
```suggestion
This module exports the `CSSStyleDeclaration` [interface wrapper API](https://github.com/jsdom/webidl2js#for-interfaces) generated by [webidl2js](https://github.com/jsdom/webidl2js). Unlike the default export, `CSSStyleDeclaration` constructors installed by the webidl2js wrapper do _not_ support construction, just like how they actually are in browsers. Creating new `CSSStyleDeclaration` objects can be done with the [`create`](https://github.com/jsdom/webidl2js#createglobalobject-constructorargs-privatedata) method of the wrapper.
```

README.md
ExE-Boss: For future reference, you can do that by using more than 3 ````​```​```` characters:

`````markdown
````suggestion
This module default-exports the `CSSStyleDeclaration` interface constructor, with the change that it can be constructed with an optional `onChangeCallback` parameter. Whenever any CSS property is modified through an instance of this class, the callback (if provided) will be called with a string that represents all CSS properties of this element, serialized. This allows the embedding environment to properly reflect the style changes to an element's `style` attribute.

Here is a crude example of using the `onChangeCallback` to implement the `style` property of `HTMLElement`:
```js
const CSSStyleDeclaration = require('cssstyle');

class HTMLElement extends Element {
  constructor() {
    this._style = new CSSStyleDeclaration(newCSSText => {
      this.setAttributeNS(null, "style", newCSSText);
    });
  }
  get style() {
    return this._style;
  }
  set style(text) {
    this._style.cssText = text;
  }
}
```
````
`````

File: index.js (hunk 11 above)
Comment (ExE-Boss) [reply index.js:15]: Well, the existing version of the constructor supports passing `null`.

File: scripts/download_latest_properties.js [hunk 17]
```diff
@@ -69,7 +69,7 @@ request(url, function(error, response, body) {
     out_file.write('/*\n *\n * https://www.w3.org/Style/CSS/all-properties.en.html\n */\n\n');
     out_file.write(
       'module.exports = new Set(' +
-        JSON.stringify(CSSpropertyNames.map(camelToDashed), null, 2) +
+        JSON.stringify(CSSpropertyNames.map(idlAttributeToCSSProperty), null, 2) +
```
TimothyGu, line 72: Is this safe to do? `idlAttributeToCSSProperty` takes a second parameter

scripts/generate_implemented_properties.js
```diff
@@ -4,12 +4,18 @@ const fs = require('fs');
...
+const webkitPropertyName = /^webkit[A-Z]/;
 const dashedProperties = fs
   .readdirSync(path.resolve(__dirname, '../lib/properties'))
-  .filter(propertyFile => propertyFile.substr(-3) === '.js')
-  .map(propertyFile => camelToDashed(propertyFile.replace('.js', '')));
+  .filter(propertyFile => path.extname(propertyFile) === '.js')
+  .map(propertyFile => {
+    return idlAttributeToCSSProperty(
+      path.basename(propertyFile, '.js'),
+      webkitPropertyName.test(propertyFile)
```
TimothyGu, line 16: Could you add `/* lowercaseFirst = */` as in the other place?

jsakas:
> Looks good as far as I’m concerned. I’ll leave @jsakas to do a final pass and merge however.

Thanks guys for all your work here. Diving into this PR this week.

ExE-Boss:
🔔 @jsakas @TimothyGu

cdoublev:
Hello,

I created #140 before truly understanding this morning what this PR was about. The initial goal of my PR was to convert values ​​passed to an interface of CSSStyleDeclaration into CSSOMString, when that interface expected such type. But this PR already handles that.

My PR fixes many other issues and implements support for several CSS properties and value types. So I allowed myself to rebase my PR on this one. I only kept my tests related to the conversion to CSSOMString.

I would like to add a few notes here, some of which I have already made in other places.

**1/** I'm not sure I understand why `lib/properties.js` should be generated (and transformed by `babel`). In my PR, I removed this build step because it prevented me from using a "higher order setter" for CSSOMString conversion, but also to handle globally (for all properties) empty string and CSS wide keywords as an initial parsing step, before passing the CSS value to the CSS property setter.

**2/** The way CSS values ​​are parsed in this library does not conform to the CSSOM specification and especially the CSS Syntax specification, ie. a "global" parsing step rather than lexing (tokenizing), consuming tokens for parsing, and serializing tokens. A "global" parse/serialize strategy is easier to implement and perhaps more efficient, but it may be preferable and even necessary to conform to those specifications before being forced to a large rewrite. This would also make it possible to avoid using the "higher order setter" mentioned in the first point.

**3/** This PR is open since more than 1 year. Would it be possible to have this PR and/or my PR reviewed by someone else? AS noted above, it fixes many `jsdom` and `cssdom` reported issues, an many others which have not been reported yet. My motivation for contributing to this library was to be able to use objects implementing `toString()` as CSS values, and to allow `calc()` (computed or not) as value and value component, in the `jest` tests of an animation library. The only way to get these two features now seems to fork `jsdom`, because [`HTMLElement.style` is created with `new cssstyle.CSSStyleDeclaration()`](https://github.com/jsdom/jsdom/blob/04f6c13f4a4d387c7fc979b8f62c6f68d8a0c639/lib/jsdom/living/nodes/ElementCSSInlineStyle-impl.js#L7). Can you see a "light" alternative workaround?

ExE-Boss:
🔔 @jsakas @TimothyGu

PR #128, opened 2021-04-23
Improve support of un-computed calc

cdoublev:
#115 adds basic support for uncomputed `calc ()` values, but only for those of type `<dimension>`, ie. a `<length-percentage>` like for the `width` property, but not for those of type `<length>`, like for the `border-width` property, or `<percentage>`.

This PR is only a draft, for several reasons.

1. There is no support for types that can be used in `calc` but which are nested within another type, such as an `<angle> `or a `<length-percentage>` (`<color-stop>`) in a `<gradient>` for the `background-image` property.

2. There is no support for `<hsla?()>` because its values are currently computed (see next point) to `<rgba?()>`.

3. Ideally, `calc` should be computed when possible (see #101 ) to conform to the specification and browsers behavior.

4. Some recent language features are used like array destructuring, and other features like `var`. So I'm not sure of the targeted environments that should be supported.

    > The primary use case is for testing browser code in a Node environment.

    Latest and stable versions, I guess?

5. `<rgba?()>` and `<hsla?()>` should be able to support a space as a separator for values ​​other than the alpha channel (note that it would complicate the task of their parser), and a `/ `to separate the latter.

This draft provides support for properties **directly receiving** (ie. not nested in another type) a value of type `<integer>`, `<number>`, `<color>` (only `<rgba? ()>` ), `<length>`, `<percentage>` (`font-stretch` is an example but its values ​​are not parsed by `cssstyle`), `<angle>` (`offset-rotate` is an example but same as `<percentage>`).

This draft also fixes the two following issues:
- in `rgba()`, if the alpha value is not supplied (eg. `rgba(0, 0, 0)`), it is currently considered as invalid, because the parser checks for 4 arguments
- `rgba(0, 0, 0, 50%)` or `rgba(0%, 0%, 0%, 50%)` output `rgb(0, 0, 0)` because `50` is not divided by 100

### DIFF
--- a/lib/CSSStyleDeclaration.test.js
+++ b/lib/CSSStyleDeclaration.test.js
@@ -550,7 +550,47 @@ describe('CSSStyleDeclaration', () => {
 
   test('supports calc', () => {
     const style = new CSSStyleDeclaration();
+
+    // Integer
+    style.setProperty('counter-increment', 'calc(1 + 1)');
+    expect(style.getPropertyValue('counter-increment')).toEqual('calc(1 + 1)');
+
+    // Number
+    style.setProperty('opacity', 'calc(0.5 * 2)');
+    expect(style.getPropertyValue('opacity')).toEqual('calc(0.5 * 2)');
+
+    // Dimension (length or percentage)
     style.setProperty('width', 'calc(100% - 100px)');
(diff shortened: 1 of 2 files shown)


cdoublev:
Superseded by #140.

PR #130, opened 2021-04-25
fix: convert value to DOMString

### LINKED ISSUE #129: Value as is not converted to String

```
class Prop {
  toString() {
    return 'red'
  }
}

```

cdoublev:
Fix #129.


cdoublev:
Unfortunately, despite my successive attemps, this PR will not be enough in all cases. Eg. with `style.border-style = { toString: () => 'italic' }`, the setter for `border-style` will not validate the value.

I feel like this package (and the issue related to this PR) requires a pretty large overhaul. IMO `parsers.valueType` could be removed, as well as `isValid`, which are almost always equivalent to `parseType(val) !== undefined`, and each property should be associated to a parsing function run by `style.setProperty` instead of being able to "override" its setter.

If any of the maintainers think spending more time on this is pointless because of a lack of their time, please let me know!

cdoublev:
Finally found a way to fix `style.borderStyle = {toString: () => 'solid'}`.

But one notable change is that implemented properties are no longer parsed and generated with `babel`. I believe this was only required by `cssom` in order to be run in a browser. I didn't bother to install `jsdom` to run its tests with this change.

**Why do I need this problem fixed?**

I maintain a JS library for animating HTML elements. In my `jest` tests, which depends on `cssstyle` via `jsdom`, many different CSS properties have their value evaluated. In summary, I want to make a change that transforms the final statement `animated.style = stringValue` into `animated.style = { toString: () => stringValue }`, but that makes those tests failing.

I've tried redefining the animated property descriptor, so that I don't have to wait for a fix to this package and a new version of `jsdom` and` jest`:

```js
test('animated.style.property = objectWithToString', () => {
    class Prop {
        toString() {
            return 'red'
        }
    }
    function createSetter(initialSetter) {
        return function set(v) {
            return initialSetter.bind(target.style)(String(v))
        }
    }
    const target = document.createElement('div')
    const descriptor = Object.getOwnPropertyDescriptor(CSSStyleDeclaration.prototype, 'backgroundColor')
    const _descriptor = { ...descriptor, set: createSetter(descriptor.set) }

    // descriptor.set = createSetter(descriptor.set) // Doesn't work
    Object.defineProperty(CSSStyleDeclaration.prototype, 'backgroundColor', _descriptor)

    target.style.backgroundColor = new Prop

    // expect(target.style.backgroundColor = new Prop).toBe('red') // Doesn't work
    expect(target.style.backgroundColor).toBe('red')
})
```

But for some reason the first test run is successful, but not the following ones.

cdoublev:
Superseded by #140.

PR #133, opened 2021-05-03
fix: handle outline style/width

cdoublev:
Fix [#2999](https://github.com/jsdom/jsdom/issues/2999).


cdoublev:
Closing in favor of a full rewrite. See #140.

PR #141, opened 2021-07-04
keep property definition using calc and var

ericminio:
Hi,
I would like to keep the property's definition for top, left, and width when using calc and var.
Not interested in computed actual values.
Cheers


ericminio:
Included color focus from PR#127 from @snowystinger

PR #150, opened 2022-04-06
Update allProperties.js

lencioni:
I did this by cloning the repo, running npm install, and then executing
the following command:

```sh
node scripts/download_latest_properties.js
```

Finally, I ran eslint autofix on the file, which mostly just reformatted
the quotes from doubles to singles.

I hope that updating this file will allow JSDOM to avoid stripping these
valid styles from style attributes on DOM nodes.


lencioni:
We've tested this change at Airbnb by using patch package and it seems to resolve some issues we were running into with this list being outdated.

PR #155, opened 2022-10-24
build: migrate to TypeScript

raveclassic:
Hey folks!

First, thanks for this amazing lib! This makes testing with jsdom a breeze.
However, I found some missing features (also referenced here and there in the tracker, like missing support for `calc`, `var` and others). I'd like to contribute and finally fix that but I faced some issues with the current state of the codebase, which I'd like to improve first, before I start.

First, the code is quite outdated (huh, 3 years without updates), we could definitely fix that by starting with adopting the codebase to the current-date best practices.

The first thing that could be done is upgrading devDependencies (newest eslint, prettier, jest etc.) and then migrating the code to TypeScript for better confidence in the future refactoring (which is required to implement `var` aka custom properties aka `--*` properties correctly).

So, this PR is the first step towards that and here's the change list:
1. upgrade eslint, prettier and jest
2. migrate all code (including repo scripts) to TypeScript
  a. I tried to be as less invasive a possible, keeping all the logic as is. This can be refactored later, easier with support of TypeScript)
  b. The prototype-patching approach was saves as TypeScript doesn't support that, but `Object.defineProperties` still work seamlessly
3. refactor codegen scripts as working with babel ast is quite clumsy and obviously unnecessary, we can do better with `ts-morph`
4. refactor `valueType` function in `parsers` module to return and ADT instead of just a type constant. This allows to keep processed value together with the type eliminating unnecessary type checks.

PS. I discovered this library when building some integration tests with `jsdom` for the code that heavily uses custom css properties (plugin for https://github.com/callstack/linaria) and found out that sandy custom css properties are not supported. It's time to fix that!

Looking forward to any feedback possible 🙏


raveclassic:
@jsakas would like to hear any feedback on this, thanks! 🙏

domenic:
Although I appreciate the enthusiasm here, in general coming in and rewriting a project in a new language is not a great way to get started on contributing to it. If you want to create your own fork of this project in a different language, and it is well-maintained, then wed consider using it in jsdom as an opaque black-box dependency. But it's not something the other people in the jsdom organization would be able to help with; we're a JavaScript project.

PR #164, opened 2023-11-08
perf: skip unnecessary access to cssText property

autologie:
While investigating slow tests in my project written with React + Chakra UI, I found out that calling the get accessor of `cssText` property takes up a large portion of total execution time. This PR is to skip such calls when possible by making `_onChange` calls conditional.

In my project, I saw ~40% improvement of the execution time (24sec → 15sec).


domenic:
Nice! It's not clear which weekend I'll have time to do another round of jsdom releases, but I am excited to merge this (after testing it with the full jsdom test suite) as soon as I do have such time.

autologie:
Hi @domenic, thank you for taking care of that.
Looking forward to getting this verified and released :)

domenic (approved):

PR #165, opened 2023-11-10
Handle empty string and null for margin and padding subparts

andrewiggins:
Per the linked issue in [jsdom](https://github.com/jsdom/jsdom/issues/2504), the following bug exists, where `dom` is a JSDom instance:

```
const elem = dom.window.document.createElement('a');

elem.style.marginTop = '10px';
console.log(elem.style.marginTop); // '10px'

elem.style.marginTop = '';
console.log(elem.style.marginTop); // Expected '', but '10px' received.
```

This PR fixes this bug by specifying that `null` and `empty string` are valid values for subparts of the properties `margin` and `padding`. I've also added tests for the behavior of setting these properties to `undefined` (value should remain as is).

Verified that this behavior (using the repro above) matches the following browser's behavior:

Chrome 119.0.6045.123
Edge 120.0.2186.2
Firefox 120.0b8
Safari 17.1 (19616.2.9.11.7)

Fixes jsdom/jsdom#2504

### DIFF
--- a/lib/CSSStyleDeclaration.test.js
+++ b/lib/CSSStyleDeclaration.test.js
@@ -402,6 +402,24 @@ describe('CSSStyleDeclaration', () => {
     expect(style.marginTop).toEqual('0.5ex');
   });
 
+  test('setting empty string to a padding or margin works', () => {
+    var style = new CSSStyleDeclaration();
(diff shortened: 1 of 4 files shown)


domenic (approved):
Thanks, the tests make this pretty easy to accept.

PR #171, opened 2024-01-28
Extend CSS color support with css-color module

### LINKED ISSUE #134: Color function values are not properly rounded

### LINKED ISSUE #135: Opaque hexadecimal color values should be resolved to rgb

### LINKED ISSUE #136: Invalid hexadecimal color value is parsed as valid

### LINKED ISSUE #174: Fails to parse "modern rgb syntax" and some other valid colors
A color in "legacy rgb syntax" with no alpha (e.g. `rgb(255, 255, 255)`) appears to be properly handled by `parseColor`.  However there are other valid color formats that are not handled.

See https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#formal_syntax

The following valid CSS colors are not parsed properly by `parseColor`:
* `rgb(255 255 255)`
* `rgb(255 255 255 / 0.5)`
* `rgb(255 255 255 / 50%)`
* `rgb(255 255 255 / none)`
* `rgb(255, 255, 255, 0.5)`
* `rgb(255, 255, 255, 50%)`
* `rgb(255 100% 100%)`
* `rgb(255 none none)`

Note all of those should also work with `rgba()`, as it has the same syntax as `rgb()` - e.g. the alpha value is optional in both.

asamuzaK:
Currently, support for CSS colors is very limited.
The [css-color](https://github.com/asamuzaK/cssColor) module covers most of the colors in CSS color module level 4.

NOTE: Colors are converted to `specified value`, not `computed value`.

NOTE2: In the test, caching seems to be working.
For this reason, I think the elapsed time displayed in test results are not so reliable.

If I add the same test, elapsed time significantly decreased.
```
  ✔ should normalize color-mix values (217.8072ms)
  ✔ should normalize color-mix values (2nd attempt) (61.3616ms)
  ✔ should normalize color-mix values (3rd attempt) (0.3769ms)
```

Fixes #134, fixes #135, fixes #136, fixes #174


asamuzaK:
PR ready.
I think we can probably delete `lib/named_colors.json` as well, but I'm not sure, so I left it as is.

domenic (approved):
Very nice, thank you!

Issue #154, opened 2022-10-13
Allow usage of `calc` for margins


coeing:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

I'd like to suggest that `TYPES.CALC` should be valid for margins as well. I used `patch-package` to use it in my project and it does not seem to make any problems.

The real change probably has to be done here: https://github.com/jsdom/cssstyle/blob/b527ed722364dc6d156487c652df100572075dee/lib/properties/margin.js#L12 

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `cssstyle@2.3.0` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/cssstyle/lib/properties.js b/node_modules/cssstyle/lib/properties.js
index 7f462f3..2af2e4e 100644
--- a/node_modules/cssstyle/lib/properties.js
+++ b/node_modules/cssstyle/lib/properties.js
@@ -1225,7 +1225,7 @@ var margin_local_var_isValid = function (v) {
   }
 
   var type = external_dependency_parsers_0.valueType(v);
-  return type === margin_local_var_TYPES.LENGTH || type === margin_local_var_TYPES.PERCENT || type === margin_local_var_TYPES.INTEGER && (v === '0' || v === 0);
+  return type === margin_local_var_TYPES.CALC || type === margin_local_var_TYPES.LENGTH || type === margin_local_var_TYPES.PERCENT || type === margin_local_var_TYPES.INTEGER && (v === '0' || v === 0);
 };
 
 var margin_local_var_parser = function (v) {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #175, opened 2024-06-26
Add calc support for margin and padding

BramMeerten:
Add support for `calc` for `margin` and `padding`. Other properties such as `width`, `top`, ... already support calc.

More properties should support `calc`, but I haven't included them all in this PR:
- `background-position: calc(100% - 0%) calc(100% - 10%)`. Not included because `parse` function splits on whitespace, but whitespace can also occur inside calc parentheses.
- `border-bottom-width` (and other "length" types): uses `parseLength` (from `parsers.js`), didn't want to change because too many usages. It could be done though, for example `parseMeasurement` already allows `calc`.

This PR should fix https://github.com/jsdom/cssstyle/issues/154

### DIFF
--- a/lib/properties/margin.js
+++ b/lib/properties/margin.js
@@ -12,6 +12,7 @@ var isValid = function (v) {
     type === TYPES.NULL_OR_EMPTY_STR ||
     type === TYPES.LENGTH ||
     type === TYPES.PERCENT ||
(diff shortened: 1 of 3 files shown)


BramMeerten:
@domenic could you please have a look at this small PR?

domenic (approved):

PR #179, opened 2024-12-20
Support nested var() and nested calc()

### LINKED ISSUE #125: allow var functions in style.color values

asamuzaK:
Fixes #125, closes #127

Based on #127. Many thanks.


asamuzaK:
PR ready.

domenic (approved):
Thank you!

Issue #184, opened 2025-02-03
Using `border-color` + `var()` doesn't work


bawjensen:
After recent changes such as #179, I wanted to see if it was possible to remove a hacky patch we've got on `cssstyle`. However, seems like it's not fully fixed, because while the parser interprets `rgba(var(...))` properly as `TYPES.VAR`, that doesn't pass the `borderColor.isValid` check. However, notably `backgroundColor.isValid` was functioning as desired. So it seems like the crux of the issue is that the border logic doesn't use `parsers.parseColor` like the background logic does.

This is largely speculation, though.

PR #185, opened 2025-02-03
Use parsers.parseColor() in border-color parsing, like background-color

bawjensen:
My attempt at fixing #184. Please feel free to use this code however ends up most useful.

### DIFF
--- a/lib/properties/borderColor.js
+++ b/lib/properties/borderColor.js
@@ -8,10 +8,7 @@ var parser = function (v) {
   if (parsed !== undefined) {
     return parsed;
   }
-  if (
-    parsers.valueType(v) === parsers.TYPES.KEYWORD &&
-    (v.toLowerCase() === 'transparent' || v.toLowerCase() === 'inherit')
-  ) {
+  if (parsers.valueType(v) === parsers.TYPES.KEYWORD && v.toLowerCase() === 'inherit') {
     return v;
   }
   return undefined;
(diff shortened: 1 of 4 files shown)


lib/properties/borderColor.js
```diff
@@ -3,23 +3,25 @@
...
 var parser = function (v) {
-  if (isValid(v)) {
-    return v.toLowerCase();
+  var parsed = parsers.parseColor(v);
+  if (parsed !== undefined) {
+    return parsed;
+  }
+  if (
+    parsers.valueType(v) === parsers.TYPES.KEYWORD &&
+    (v.toLowerCase() === 'transparent' || v.toLowerCase() === 'inherit')
```
asamuzaK, line 13: I think we don't need `(v.toLowerCase() === 'transparent'`.
It's already resolved by `css-color`.

To make sure, add test like below somewhere around test/parsers.js#L64
```
  it('returns color for transparent keyword', () => {
    let input = 'transparent';
    let output = parsers.valueType(input);

    assert.strictEqual(output, parsers.TYPES.COLOR);
  });
```

And before test/parsers.js#L248 i.e. `it.todo('Add more tests')`
```
  it('should output transparent keyword', () => {
    let input = 'transparent';
    let output = parsers.parseColor(input);

    assert.strictEqual(output, 'transparent');
  });
```

bryan-codaio: I had effectively copy/pasted this from the `background-color` logic. Should I update that too, to keep in sync? (I've proactively updated it, but as a separate commit that should be easy to revert if needed)

bryan-codaio: thanks for the review!

asamuzaK (approved):
LGTM

Issue #189, opened 2025-04-01
Trying to parse undefined will cause error in cssColor dependency's resolveColor


kasimir-dka:
https://github.com/jsdom/cssstyle/blob/4533e6f35305dd154d693534fb4cf01dacdb72e9/lib/parsers.js#L268

In `parseColor` we test if value is empty string or null, otherwise we continue to use `resolveColor`: 
```js
  if (type === exports.TYPES.NULL_OR_EMPTY_STR) {
    return val;
  }
  if (/^[a-z]+$/i.test(val) && type === exports.TYPES.COLOR) {
    return val;
  }
  var res = resolveColor(val, {
    format: 'specifiedValue',
  });
``` 
If `val` is undefined, this will cause an error in `resolveColor`

We can prevent this by adding a test for undefined:
```js
  if (typeof val === undefined || type === exports.TYPES.NULL_OR_EMPTY_STR) {
    return val;
  }
```

PR #190, opened 2025-04-04
Add a check for undefined val for parseColor

kasimir-dka:
Fixes issue https://github.com/jsdom/cssstyle/issues/189

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -267,7 +267,7 @@ exports.parseString = function parseString(val) {
 
 exports.parseColor = function parseColor(val) {
   var type = exports.valueType(val);
-  if (type === exports.TYPES.NULL_OR_EMPTY_STR) {
+  if (typeof val === 'undefined' || type === exports.TYPES.NULL_OR_EMPTY_STR) {
     return val;
   }
   if (/^[a-z]+$/i.test(val) && type === exports.TYPES.COLOR) {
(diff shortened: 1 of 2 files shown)


domenic:
Can you explain why you are passing undefined to this function in the first place?

asamuzaK:
I've looked into it a bit deeper and it seems there's nothing wrong with returning undefined early.
Forget my previous comment.
Sorry for a noise.

domenic:
I guess I am not excited to merge this until we can get a test in the upstream jsdom repository that this solves.

I also wonder if this is just one instance of https://github.com/jsdom/cssstyle/issues/129.

asamuzaK:
The end result will be the same, but the following steps may be more consistent with other features?

``` javascript
exports.parseColor = function parseColor(val) {
  var type = exports.valueType(val);
  if (type === exports.TYPES.NULL_OR_EMPTY_STR) {
    return val;
  }
  if (type !== exports.TYPES.COLOR) {
    return undefined;
  }
  if (/^[a-z]+$/i.test(val)) {
    return val;
  }
  var res = resolveColor(val, {
    format: 'specifiedValue',
  });
  if (res) {
    return res;
  }
  return undefined;
};
```

asamuzaK:
Re #129
I'm having difficulty following what the reporter of #129 is trying to say.
In particular, it's unclear whether the results written in the comments are expected values ​​or actual values.

asamuzaK:
Test summary: https://github.com/jsdom/cssstyle/issues/196#issuecomment-2824705234

PR #192, opened 2025-04-15
Update cssColor

### LINKED ISSUE #191: calc(10px + 20%) should serialize to calc(20% + 10px)
This is what remains of https://github.com/jsdom/jsdom/issues/1332.

asamuzaK:
Fix #191


domenic (approved):
I didn't realize this package was getting its calc() implementation from a "color"-related package, but I guess it works :)

PR #194, opened 2025-04-19
Update parseUrl function

### LINKED ISSUE #103: update url parsing logic to match current browsers
it would be nice if url parsing worked like latest firefox/chrome.

for instance, running the following snippet in those browsers:
```
(() => {
  var div = document.createElement('div');
  div.style.backgroundImage = '   url(http://some/url/here1.png)   , URL(2)  ';
  console.log(`backgroundImage:${div.style.backgroundImage};`);
})();
```

prints:
```
backgroundImage:url("http://some/url/here1.png"), url("2");
```

i've searched the code a bit, and think we can achieve this by modifying the following function:
https://github.com/jsdom/cssstyle/blob/f2db1ebef45c03b51787f5b9d00a3eb843f07787/lib/parsers.js#L210-L247

we can also add some better unit tests here:
- https://github.com/jsdom/cssstyle/blob/f2db1ebef45c03b51787f5b9d00a3eb843f07787/lib/parsers.test.js#L63-L65
- https://github.com/jsdom/cssstyle/blob/f2db1ebef45c03b51787f5b9d00a3eb843f07787/lib/CSSStyleDeclaration.test.js#L377-L385

if i make the "formatted" string look like `url("foo.png")` instead of the current `url(foo.png)`, then I could see that breaking people's existing tests (but it would "match" modern browsers more closely).  If I just allow trailing/closing whitespace, and multiple urls, then that would be "more" backwards compatible I think.

I can work on this in the next few days and submit a PR.  I have quite a few ideas on how to get it to work.

Thanks!

asamuzaK:
Fix #103

* Handle escaped char.
* Output URL string with quotes.
  Although the unquoted output is valid, all major browsers appear to quote the output.

This is a BREAKING CHANGE

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -33,7 +33,7 @@ var lengthRegEx = new RegExp(
 );
 var percentRegEx = new RegExp(`^${NUMBER}%$`);
 var angleRegEx = new RegExp(`^${NUMBER}(?:deg|g?rad|turn)$`);
-var urlRegEx = /^url\(\s*([^)]*)\s*\)$/;
+var urlRegEx = /^url\(\s*((?:[^)]|\\\))*)\s*\)$/;
 var stringRegEx = /^("[^"]*"|'[^']*')$/;
 var varRegEx = /^var\(|(?<=[*/\s(])var\(/;
 var calcRegEx =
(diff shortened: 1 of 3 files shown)


domenic:
> This is a BREAKING CHANGE

In what way is this breaking? It feels more like a bug fix to me.

domenic (approved):

asamuzaK:
From this https://github.com/jsdom/cssstyle/issues/103#issue-498260422

> if i make the "formatted" string look like url("foo.png") instead of the current url(foo.png), then I could see that breaking people's existing tests (but it would "match" modern browsers more closely).

domenic:
Hmm. Historically in the jsdom project we've treated these kinds of spec-conformance fixes as bug fixes. I agree peoples' tests might be affected, but I think it's best to just stick to semantic versioning in these cases.

asamuzaK:
>  I think it's best to just stick to semantic versioning in these cases.

Agreed.

PR #197, opened 2025-04-22
Fix valueType and add more tests

asamuzaK:


domenic (commented):
This looks good to me.

However can you state more about what issue this is fixing at the jsdom level? Does it help with any of the tests added in https://github.com/jsdom/jsdom/pull/3871 ? Should we do a new release and roll jsdom soon?

asamuzaK:
For now, there's no point other than to increase test coverage.
There seems to be a lot of unnecessary processing in valueType(), so I wanted to write tests before working on it.

> Should we do a new release and roll jsdom soon?

I don't think we need to release the next version of cssstyle right now.
When do you plan to release the next jsdom?
I think it would be sufficient to release next cssstyle at that time.

PR #208, opened 2025-05-10
Organize files

### LINKED ISSUE #70: "webkit-*" vs "-webkit-*"
It seems there's a problem with the webkit prefixes (leading "-").
I think it should be ```-webkit-transform```, not ```webkit-transform```.

The following code shows the issue:

```
const
  cssstyle = require('cssstyle'),
  style = new cssstyle.CSSStyleDeclaration();

style.cssText = 'background-color: red; -webkit-transform: scale(2);'

// Be aware: Using "-webkit-transform" with leading "-"
console.log('background-color: ' + style.backgroundColor)  // working => "red"
console.log('-webkit-transform: ' + style.webkitTransform) // not working => "" <---------- !!!!


// Be aware: Using "webkit-transform" now, without leading "-"
style.cssText = 'background-color: green; webkit-transform: scale(3);'

console.log('background-color: ' + style.backgroundColor)  // working => "green"
console.log('webkit-transform: ' + style.webkitTransform) // working => "scale(3)"
```

### LINKED ISSUE #124: 'border' setter does not validate new value
Code:
```javascript
const { CSSStyleDeclaration } = require('cssstyle');

// NOTE: Throws exception
style.border = undefined;

// NOTE: Works fine
style.borderWidth = undefined;
```

Exception: 
```
TypeError: Cannot read property 'toString' of undefined
    at CSSStyleDeclaration.set [as border] (/Users/andrey.churkin/Documents/git/dx-grid-error/node_modules/cssstyle/lib/properties.js:419:11)
    at Object.<anonymous> (/Users/andrey.churkin/Documents/git/dx-grid-error/index_css.js:4:14)
    at Module._compile (internal/modules/cjs/loader.js:1075:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1096:10)
    at Module.load (internal/modules/cjs/loader.js:940:32)
    at Function.Module._load (internal/modules/cjs/loader.js:781:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

Look [here](https://github.com/jsdom/cssstyle/blob/master/lib/properties/border.js#L17).

### LINKED ISSUE #129: Value as is not converted to String

### LINKED ISSUE #200: [Meta] code quality
* [ ] #206
  * [ ] #201 
  * [ ] #207
* [ ] #202
* [ ] #205
* [ ] #203
* [ ] #204

### LINKED ISSUE #201: Move `parsers.shorthandGetter()`, `parsers.shorthandSetter()`, `parsers.implicitSetter()` and `parsers.subImplicitSetter()` to `CSSStyleDeclaration()`

### LINKED ISSUE #202: Remove `property.isValid()` or replace them to more effective function

### LINKED ISSUE #203: Get rid of `TYPES.NULL_OR_EMPTY_STRING` and `TYPES.UNDEFINED`

### LINKED ISSUE #204: Convert `var` to `const` and `let`

### LINKED ISSUE #205: Avoid returning undefined for falsy results

### LINKED ISSUE #206: Organe files and directories

### LINKED ISSUE #207: Camelize-related files into one

asamuzaK:
* Created `utils/camelize.js`
* Put generated files under `generated/`
* Switched to `const` and `let`
* Updated code style rules through prettier
* Added `@domenic/eslint-config` rules
* Updated CSSStyleDeclaration class to match the latest spec
  see https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface
  * `cssText` getter:
    * Returns empty string if computed flag is set
  * `cssText` setter:
    `removeProperty()`:
    `setProperty()`:
      * Throws if readonly flag is set
  * `parentRule()` getter:
    * Returns parent `CSSRule` if associated

  Note that the above flags and methods needs additional fixes on `jsdom` side
  * `item()`:
    * Throws if no argument given
      see also https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/item
  * `-webkit-*` prefixed properties:
    * Adds Camel Cased and Pascal Cased properties, e.g.) `style['-webkit-foo']`, `style.webkitFoo` and `style.WebkitFoo` are added
* Rewritten `properties/*.js` files
  * Unified each file to export three functions: parser (`exports.parse`), validator (`exports.isValid`), and descriptor (`exports.definition`)
  * Removed deprecated `azimuth` and some deprecated `webkit` prefixed properties
    see also https://developer.mozilla.org/en-US/docs/Web/CSS/WebKit_Extensions
* Renamed test files -> *.test.js and added tests

Fixes #200, Fixes #201, Fixes #202, Fixes #203, Fixes #204, Fixes #205,
Fixes #206, Fixes #207, Fixes #70, Fixes #124, Fixes #129,
Fix jsdom issues: jsdom/jsdom#3878, jsdom/jsdom#3833
Also we can close #120, I think.


asamuzaK:
@domenic 
Almost ready.

Do you think we still need `webkitXXX.js` files under `./properties`?
Is it okay to remove them?

domenic:
I'm not sure I understand enough project architecture to know exactly what those files are currently doing.

We need to support this part of the spec: https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface:~:text=fontSize%20IDL%20attribute.-,For%20each%20CSS%20property%20property,exceptions%20thrown%20must%20be%20re%2Dthrown.,-For%20example%2C%20if . Do those files help with that?

lib/properties/bottom.js
```diff
@@ -1,14 +1,20 @@
...
-var parseMeasurement = require('../parsers').parseInheritingMeasurement;
+const parseMeasurement = require("../parsers").parseInheritingMeasurement;
 
 module.exports.definition = {
-  set: function (v) {
-    this._setProperty('bottom', parseMeasurement(v));
+  set(v) {
+    if (v === undefined) {
+      return;
+    }
```
domenic, line 9: My reading of the spec is that `undefined` should behave as the string `"undefined"`. In most cases this will fail to parse and thus do nothing, but in some cases it probably has an effect?

asamuzaK:
> I'm not sure I understand enough project architecture to know exactly what those files are currently doing.
> 
> We need to support this part of the spec: https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface:~:text=fontSize%20IDL%20attribute.-,For%20each%20CSS%20property%20property,exceptions%20thrown%20must%20be%20re%2Dthrown.,-For%20example%2C%20if . Do those files help with that?

Let's keep webkitFoo files for now.
I'll file new issue for that.

lib/properties/bottom.js
asamuzaK: Done

asamuzaK:
PR ready.

asamuzaK:
Back to draft.
I'll do one more work to fix #129

asamuzaK:
Sorry, found a bug.

asamuzaK:
It ended up being a fairly large fix, but shouldn't be a breaking change.
Please review.

asamuzaK:
> The only things I would like to request is more details on "Updated CSSStyleDeclaration class to match the latest spec" and "Rewritten properties/*.js files".

I think I can have them ready on Saturday, or at the latest by Sunday noon (JST), is that okay?

> Port the tests to web platform tests format, but write them in a way that they can be run with a small wrapper like https://github.com/jsdom/whatwg-url/blob/414f17a3459b0872baee7a2b77e23953b8a5ccd9/test/web-platform.js#L159-L188 .

Interesting.
I'll take a look at it later.

PR #215, opened 2025-06-08
fix background-* spelling

### LINKED ISSUE #214: Background properites misnamed as "backgound" (missing the 'r')
See: https://github.com/jsdom/cssstyle/blob/7ed8c6cf29bd1fcf53af19cc5726a950b7950e30/lib/properties/backgroundColor.js#L25

I suspect this may have been copy-pasted into other background properties.

fshowalter:
Closes #214

### DIFF
--- a/lib/properties/backgroundAttachment.js
+++ b/lib/properties/backgroundAttachment.js
@@ -19,7 +19,7 @@ module.exports.definition = {
     v = parsers.prepareValue(v, this._global);
     if (parsers.hasVarFunc(v)) {
       this._setProperty("background", "");
-      this._setProperty("backgound-attachemnt", v);
+      this._setProperty("background-attachemnt", v);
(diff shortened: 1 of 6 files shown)


asamuzaK:
Thanks for the catch.
Please add regression test at the end of `test/CSSDeclaration.test.js`:
``` js
describe("regression test for https://github.com/jsdom/cssstyle/issues/214", () => {
  it("should return value for each property", () => {
    const key = "background-color";
    const camel = "backgroundColor";
    const value = "var(--foo)";
    style[key] = value;
    assert.strictEqual(style[key], value);
    style[key] = null;
    style[camel] = value;
    assert.strictEqual(style[camel], value);
  });
});
```

fshowalter:
Done.

domenic (approved):

Issue #212, opened 2025-06-08
Regression with font family serialization


domenic:
Most of the tests in https://github.com/web-platform-tests/wpt/blob/master/css/cssom/font-family-serialization-001.html are failing as of https://github.com/jsdom/cssstyle/commit/79c88c407ee5da7f4fd9849725c9e9cc04b1a2c8.

@asamuzaK

PR #220, opened 2025-06-21
Add `generic()` font-family keyword support

asamuzaK:
Follow up for #212

This PR is made on top of #218
So we can close #218 too

### DIFF
--- a/test/properties.test.js
+++ b/test/properties.test.js
@@ -1395,6 +1395,26 @@ describe("font", () => {
     testPropertyValue("font-family", "Hawaii 5-0, sans-serif", "");
   });
 
+  it("font-family should set / get family values", () => {
+    testPropertyValue("font-family", "generic(fangsong)", "generic(fangsong)");
+  });
+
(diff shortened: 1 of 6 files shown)


test/properties.test.js
```diff
@@ -1395,6 +1395,26 @@ describe("font", () => {
...
+  it("font-family should set / get family values", () => {
+    testPropertyValue("font-family", "generic(khmer-mul)", "generic(khmer-mul)");
+  });
+
+  it("font-family should set / get family values", () => {
+    testPropertyValue("font-family", "generic(nastaliq)", "generic(nastaliq)");
+  });
+
+  it("font-family should not set / get invalid family values", () => {
+    testPropertyValue("font-family", "generic(foo)", "");
```
domenic, line 1415: Is there spec support for this? Chrome and Firefox return back `generic(foo)` when I set it.

domenic: Sorry, I made a typo while testing.

domenic (approved):

PR #223, opened 2025-06-24
Fix margin handlers

### LINKED ISSUE #222: Negative values invalid for `marginBottom`?
I see `restrictToPositive` is set to true for `marginBottom` (and the tests assert that the value cannot be negative) - is this correct?

https://github.com/jsdom/cssstyle/blob/ba349623a94d5d1fc6527c770e6684acc82bac44/lib/properties/marginBottom.js#L6

https://github.com/jsdom/cssstyle/blob/ba349623a94d5d1fc6527c770e6684acc82bac44/lib/parsers.js#L230

https://github.com/jsdom/cssstyle/blob/ba349623a94d5d1fc6527c770e6684acc82bac44/test/properties.test.js#L815-L817

I understood that its value can be positive, zero, or negative (see [https://developer.mozilla.org/en-US/docs/Web/CSS/margin-bottom](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-bottom#:~:text=Its%20value%20can%20be%20positive%2C%20zero%2C%20or%20negative)).

asamuzaK:
Fix #222

### DIFF
--- a/lib/properties/margin.js
+++ b/lib/properties/margin.js
@@ -5,7 +5,7 @@ const parsers = require("../parsers");
 const positions = ["top", "right", "bottom", "left"];
 
 module.exports.parse = function parse(v) {
-  const val = parsers.parseMeasurement(v, true);
+  const val = parsers.parseMeasurement(v);
   if (val) {
     return val;
   }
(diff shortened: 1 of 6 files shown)


domenic:
A web platform test in the jsdom repo would be appreciated (but not required).

PR #230, opened 2025-07-12
Raise minimum Node.js version to v20

asamuzaK:
* `.github/workflows/build.yml`
  Drop node v18 and add node v22 as `lts/*`
* `package.json`
  Bump `engines.node` to `>=20`
* `lib/properties/font.js`
  Switch to Array.prototype.toReversed()
* `scripts/generateImplementedProperties.mjs`
  * Switch to `import.meta.dirname`
  * Sort generated properties

<ins>Done below</ins>

TODO:
Fix `properties/font.js` after #227 get merged.

TBD:
Should we fix `scripts/generateImplementedProperties.mjs` too?
However, `import.meta.dirname` is not standard.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 18
           - 20
+          - lts/*
           - latest
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


domenic:
> TBD:
> Should we fix `scripts/generateImplementedProperties.mjs` too?
> However, `import.meta.dirname` is not standard.

I think we should fix it. `import.meta` is not governed by standards, so since we are running on Node.js, we can use anything that they officially support. Per https://nodejs.org/api/esm.html#importmetadirname `import.meta.dirname` is officially supported.

PR #234, opened 2025-07-21
Use css-tree

asamuzaK:
Switch to [CSSTree](https://www.npmjs.com/package/css-tree) to improve code quality.
* Add some fixes to CSSStyleDeclaration setter

Related wpts:
* to-upstream
  * css/css-syntax/parse-invalid-input.html
* css/cssom
  * serialization-CSSDeclaration-with-important.html

* css/css-color
  * light-dark-currentcolor-in-color.html
    Fails but this is expected. Need to fix getComputedStyle()

### DIFF
--- a/test/CSSStyleDeclaration.test.js
+++ b/test/CSSStyleDeclaration.test.js
@@ -2,7 +2,6 @@
 
 const { describe, it } = require("node:test");
 const assert = require("node:assert/strict");
-const CSSOM = require("rrweb-cssom");
 const { CSSStyleDeclaration } = require("../lib/CSSStyleDeclaration");
 const allExtraProperties = require("../lib/allExtraProperties");
 const allProperties = require("../lib/generated/allProperties");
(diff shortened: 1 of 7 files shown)


domenic:
It would be nice if jsdom only had one CSS parser in its dependency tree. https://github.com/jsdom/jsdom/pull/3910 is trying to replace rrweb-cssom with a better replacement. Maybe jsdom should use csstree instead as well?

asamuzaK:
Ah, found https://github.com/csstools/postcss-plugins/tree/main/packages/css-syntax-patches-for-csstree
i will add one more push.

test/CSSStyleDeclaration.test.js
```diff
@@ -140,8 +139,51 @@ describe("CSSStyleDeclaration", () => {
     assert.strictEqual(style.cssText, "color: green;");
   });
 
+  it("sets internals for Element", () => {
```
domenic, line 142: These test names don't seem very accurate?

domenic (commented):
Why did you remove `@csstools/css-syntax-patches-for-csstree`? (Also, when you do add it, we should add a test which depends on that package existing.)

asamuzaK:
cssTree.lexer is very powerful and it can probably replace our isValid functions under properties/.
But thats another issue and I will take a look later.

domenic (approved):
I am a little worried that we only replaced one instance of `parseKeyword()`, and left all the existing ones alone. However, I cannot see any obvious cases where this halfway state would cause problems, so I will merge this.

I would love it if you were able to open an issue detailing your plans for replacing more and more parts of this library with css-tree. (Or maybe discussing in #235 is best.)

asamuzaK:
Could you please release a new version?
Then I can start working on fixing https://github.com/jsdom/jsdom/issues/3895 and else.

PR #245, opened 2025-09-15
Fix parsing flex-basis in shorthand flex when not a value

matmannion:
Fixes the case in the flex shorthand where the third value isn't an explicit value or dimension, which was causing a parse error when called with (e.g.) `flex: 0 0 auto`

### DIFF
--- a/lib/properties/flex.js
+++ b/lib/properties/flex.js
@@ -53,17 +53,15 @@ module.exports.parse = function parse(v, opt = {}) {
           return name;
         }
         case "Identifier": {
-          if (name === "auto") {
-            flex["flex-basis"] = name;
-            return flex;
-          } else if (name === "none") {
+          if (name === "none") {
             return {
               "flex-grow": "0",
               "flex-shrink": "0",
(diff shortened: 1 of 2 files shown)


domenic:
@asamuzaK can you take a look?

@matmannion would you be willing to send a regression test to jsdom as well? Similar to, e.g., https://github.com/jsdom/jsdom/pull/3893

asamuzaK:
We also need to fix L56 block to:
``` js
if (name === "none") {
  return {
    "flex-grow": "0",
    "flex-shrink": "0",
    "flex-basis": "auto"
  };
}
flex["flex-basis"] = name;
return flex;
```

And L126 block to:
``` js
case "Identifier": {
  flex["flex-basis"] = val2.name;
  break;
}
```

asamuzaK:
LGTM

heath-freenome:
Sweet! When can we get this released?

domenic:
jsdom releases usually happen on weekends when I am free (which is not every weekend).

PR #250, opened 2025-11-06
Fix: border Hash colorParse

cplepage:
Fixes https://github.com/jsdom/cssstyle/issues/251


cplepage:
@domenic using hash color with `border` property is already broken in jsdom: https://github.com/jsdom/jsdom/actions/runs/19126666454/job/54658134773?pr=3980#step:6:3028

How do you want to proceed to fix

PR #261, opened 2025-12-13
Preparation for resolving shortcut property issues

asamuzaK:
### Summary
This PR serves as a preparation step to resolve long-standing issues with shorthand properties.
It introduces a mechanism to auto-generate CSS property definitions and refactors the codebase to utilize them.

#### 1. Introducing `propertyDefinitions.js`
- **Change:** Added a script to generate `lib/generated/propertyDefinitions.js` based on property metadata.
- **Reason:** To implement shorthand properties correctly, we need a reliable source of truth for every longhand property (initial values, syntax, inheritance, etc.). Hardcoding these manually is error-prone.
- **Future Integration with jsdom:** The newly introduced `lib/generated/propertyDefinitions.js` is designed to be the single source of truth for CSS properties. In the future, this is intended to replace `propertiesWithResolvedValueImplemented` in jsdom's `lib/jsdom/living/helpers/style-rules.js`, allowing jsdom to delegate property metadata management entirely to `cssstyle`.

#### 2. Splitting Functions and Clarifying Roles
- **Change:** Decomposed the monolithic normalization logic into smaller, discrete functions.
- **Reason:** To explicitly define the role of each function (separation of concerns). By isolating the logic for different value types and validation steps, the code becomes more readable and easier to maintain. This modularity is essential for handling the complexity of shorthand property expansion in the next steps.

#### 3. Refactoring with Constants
- **Change:** Replaced string literals (e.g., property names) with constants.
- **Reason:** To prevent bugs caused by typos and to improve code maintainability across the parser and normalizer.

### Effect
While the direct effects of this PR are limited regarding the overall shorthand behavior, it explicitly resolves the following test case:
- `web-platform-tests/to-upstream/css/cssom/style-border-shorthand-var.html`

Once merged, this PR establishes the necessary data structures and infrastructure, allowing us to implement the complex logic for expanding shorthands into longhands in a follow-up PR with high reliability and low maintenance cost.

### DIFF
--- a/lib/normalize.js
+++ b/lib/normalize.js
@@ -1294,52 +1294,23 @@ const processShorthandProperties = (shorthands) => {
 };
 
 /**
- * Updates longhand properties for all 4 border positions (top, right, bottom, left).
+ * Updates the longhand properties map with a new property item.
+ * If a property with normal priority already exists, it will be overwritten by the new item.
+ * If the existing property has "important" priority, it will not be overwritten.
  *
- * @param {Map} properties - The map of properties to update.
- * @param {string} namePart - The property name prefix (e.g., "border").
- * @param {string} linePart - The property line suffix (e.g., "width", "style", "color").
- * @param {string|Array} value - The value to set (string or array for position-based values).
- * @param {string} priority - The priority of the property.
+ * @param {string} property - The CSS property name.
+ * @param {Object} item - The property item object containing value and priority.
+ * @param {Map} longhandProperties - The map of longhand properties to update.
  */
-const updatePositionLonghands = (properties, namePart, linePart, value, priority) => {
-  for (const position of borderPositions) {
-    const longhandProperty = `${namePart}-${position}-${linePart}`;
-    const longhandValue = Array.isArray(value) ? getPositionValue(value, position) : value;
-    const longhandItem = createPropertyItem(longhandProperty, longhandValue, priority);
-    if (properties.has(longhandProperty)) {
-      const { priority: existingPriority } = properties.get(longhandProperty);
-      if (!existingPriority) {
-        properties.delete(longhandProperty);
-        properties.set(longhandProperty, longhandItem);
-      }
-    } else {
-      properties.set(longhandProperty, longhandItem);
+const updateLonghandProperties = (property, item, longhandProperties) => {
+  if (longhandProperties.has(property)) {
+    const { priority: longhandPriority } = longhandProperties.get(property);
+    if (!longhandPriority) {
+      longhandProperties.delete(property);
+      longhandProperties.set(property, item);
     }
-  }
-};
-
-/**
- * Updates longhand properties based on a values object.
- *
- * @param {Map} properties - The map of properties to update.
- * @param {Object} values - The object containing property keys and values.
- * @param {string} priority - The priority of the properties.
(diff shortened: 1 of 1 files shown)


domenic:
Can you add a more detailed description of this PR? What changes, why are those changes good, does this fix any bugs currently or is it only for use in a follow-up, etc.? Why do we need a new generated properties list and how does it differ from the other generated lists? What specs does this implement? That sort of thing.

lib/normalize.js
```diff
@@ -212,648 +298,792 @@ const replaceBorderShorthandValue = (value, shorthandValue, opt = {}) => {
...
-        priority: ""
-      };
-    borderProperties.set(lineProperty, lineItem);
+  const borderProps = new Map([[BORDER, borderItems.get(BORDER)]]);
+  for (const line of borderLines) {
+    const lineProperty = `${BORDER}-${line}`;
+    const lineItem =
+      borderItems.get(lineProperty) ??
+      properties.get(lineProperty) ??
+      createPropertyItem(lineProperty);
```
domenic: This can be replaced by getPropertyItem()

lib/normalize.js
```diff
@@ -862,272 +1092,168 @@ const generateBorderPositionShorthand = (items, property, prior) => {
...
-    const [property, item] =
-      generateBorderLineShorthand(lineColorPriorItems, "border-color", "important") ?? [];
-    if (property && item) {
-      properties.set(property, item);
-    }
+const borderCollections = {
+  [WIDTH]: {
+    shorthand: BORDER_WIDTH,
+    generator: generateBorderLineShorthand,
+    items: new Map(),
```
domenic: This module-level mutable state is worrying. Will it cause problems if multiple jsdom instances use it? At the very least it needs good documentation.

lib/normalize.js
```diff
@@ -1149,48 +1275,193 @@ const prepareBorderShorthands = (properties) => {
...
+            // fallback for safety, though lineProperty from border.parse keys
+            // should have hyphen
+            namePart = lineProperty;
+            linePart = "";
+          }
+          const lineValue = lineItems[lineProperty];
+          for (const position of borderPositions) {
+            const longhandProperty = `${namePart}-${position}-${linePart}`;
+            const longhandItem = createPropertyItem(longhandProperty, lineValue, priority);
+            if (longhandProperties.has(longhandProperty)) {
```
domenic: This pattern (lines 1346-1354) appears multiple times. Can we extract it into a helper?

asamuzaK:
For now, I've addressed two of the points you raised in the review.
Please confirm if this is the right direction.
If so, I'll move on to the rest.

lib/normalize.js
asamuzaK: https://github.com/jsdom/cssstyle/pull/261/commits/66437d1d888a3124529363599d5021d30a693bd7

Please take a look.

lib/normalize.js
asamuzaK: https://github.com/jsdom/cssstyle/pull/261/commits/6507ee383221a740a2861604b816180b0aef0eca

Please take a look.

asamuzaK: https://github.com/jsdom/cssstyle/pull/261/commits/15cd48cfdee3d26b21a9b65e1f6fe847cd9046a2
Is this what you're saying?

PR #269, opened 2025-12-26
Improve asciiLowercase/asciiUppercase performance

asamuzaK:
See https://github.com/jsdom/jsdom/commit/1cd029efb72314840c46730d72f9684c60635c54


domenic:
I should create a whatwg-infra package or something to deduplicate these across the jsdom ecosystem... with benchmarks...

But for now let's just merge this :)

PR #270, opened 2025-12-26
Refactor and optimize parsers.js

asamuzaK:
## **Summary**

This PR refactors parsers.js to improve performance, maintainability, and code readability. The changes include optimizing lookup operations, eliminating redundant object creation in loops, reducing code duplication, and standardizing coding styles.

### **1\. Performance Optimizations**

* **O(1) Lookups**: Converted GLOBAL\_KEYS (formerly GLOBAL\_KEY) and SYS\_COLORS (formerly SYS\_COLOR) from Arrays to Sets. This changes the complexity of keyword checks from O(N) to O(1).  
* **Remove Dynamic RegExp**: Replaced the dynamic creation of new RegExp inside loops within parsePropertyValue with string manipulation methods (slice). This reduces the overhead of compiling regular expressions on every iteration.

### **2\. Refactoring & Maintainability**

* **Common Numeric Parser**: Introduced a helper function parseNumericValue to centralize the parsing logic for numbers, lengths, percentages, and angles. This reduces code duplication across parseNumber, parseLength, parsePercentage, etc.  
  * **Design Note**: The validation logic is passed as an anonymous callback function rather than being hardcoded in the helper. This design was chosen because the handling of unitless 0 varies across CSS types (e.g., allowed for \<length\>, but invalid for others such as \<time\>). Defining the condition explicitly in the caller function maintains clarity and flexibility for type-specific rules.  
* **AST Type Constants**: Introduced AST\_TYPES to manage AST node types (e.g., "Function", "Identifier") as constants, eliminating magic strings and reducing the risk of typos.  
* **Module Export Style**: Refactored the module to define functions using const and export them via a single module.exports block at the end of the file.

### **3\. Documentation**

* **JSDoc**: Added comprehensive JSDoc comments to all exported functions to provide type information and improve the developer experience.

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -326,33 +326,32 @@ const parsePropertyValue = (prop, val, opt = {}) => {
             const raw = items.length === 1 ? val : css;
             // Remove "${name}(" from the start and ")" from the end
             const itemValue = raw.slice(name.length + 1, -1).trim();
-
             if (name === "calc") {
               if (children.length === 1) {
                 const [child] = children;
                 if (child.type === AST_TYPES.NUMBER) {
                   parsedValues.push({
                     type: AST_TYPES.CALC,
-                    name: "calc",
                     isNumber: true,
                     value: `${parseFloat(child.value)}`,
+                    name,
                     raw
                   });
                 } else {
                   parsedValues.push({
                     type: AST_TYPES.CALC,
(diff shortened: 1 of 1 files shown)


domenic (approved):
Looks great!

Just to confirm, this does not change any observable behavior or fix any bugs?

PR #273, opened 2025-12-29
Enhance property exports and implement AST-based resolvers

asamuzaK:
### Description:
This PR improves the internal architecture of `cssstyle` by refining how property modules export their metadata and how values are resolved via AST types. It also ensures better compliance with Web Platform Tests and streamlines the value processing logic.

### Key Changes:
* **Property Name Exports:** Each property module now explicitly exports its property name as `module.exports.property`. This replaces previously hardcoded strings (e.g., `"background"`) with a more maintainable, reference-based approach for property names.
  * NOTE: Planning to apply this to `normalize.js` in a separate follow-up PR.
* **New AST Resolvers in `parsers.js`:** Added a set of helper functions (resolvers) that dispatch values to the appropriate parser based on their AST (Abstract Syntax Tree) type. This logic has been applied across the property files in `lib/properties/*.js` to streamline value processing.

### Additional Improvements:
* **Refactored Priority Calculation:** As a related cleanup, the logic for determining the `priority` argument in property setters has been simplified and unified. This reduces code duplication and ensures consistent behavior across different properties.

### Impact:
* **Supersedes #272:** This PR covers the objectives of #272 with a more robust implementation and should be considered its replacement.

### Testing:
* Verified that `npm test` passes.
* Confirmed the fix for wpt:tuwpt `css/cssom/serialize-whitespace.html`.

### DIFF
--- a/lib/utils/propertyDescriptors.js
+++ b/lib/utils/propertyDescriptors.js
@@ -2,18 +2,54 @@
 
 const parsers = require("../parsers");
 
-exports.getPropertyDescriptor = function getPropertyDescriptor(property) {
-  return {
-    set(v) {
-      v = parsers.parsePropertyValue(property, v, {
-        globalObject: this._global
+const { AST_TYPES } = parsers;
+
+const getPropertyDescriptor = (property) => ({
+  set(v) {
+    const value = parsers.prepareValue(v, this._global);
+    if (parsers.hasVarFunc(value)) {
+      this._setProperty(property, value);
+    } else {
+      const parsedValue = parsers.parsePropertyValue(property, v, {
+        globalObject: this._global,
+        inArray: true
       });
-      this._setProperty(property, v);
-    },
-    get() {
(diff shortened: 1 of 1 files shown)


asamuzaK:
`getPropertyDescriptor()` is a function that returns generic getters and setters for properties **not** included in `properties/*.js`.
In the original implementation, the setter only performed a simple process of parsing the CSS input and setting it as-is. In this PR, I have updated it to handle properties more appropriately (at least to a minimum extent) by utilizing `AST_TYPES`.

I had already submitted the changes to `getPropertyDescriptor()` as a separate PR #272, but since both `properties/*.js` and `utils/propertyDescriptor.js` now share the same logic of importing and using `AST_TYPES`, I decided to put them into this single PR.

domenic:
I've updated the commit description to include that and will merge after CI finishes.

As always, let me know when is a good time for a new release!

PR #275, opened 2025-12-31
Enforce string conversion to prevent TypeErrors

asamuzaK:
## Motivation
Previously, the code would throw an exception if `value.toString()` returned a non-string value.
However, this was inconsistent with browser behavior.
We now force the value into a string using a template literal.
This guarantees that the function always returns a valid string instead of causing a runtime error.

## Impact
- Refs #129
- Verified that `test/web-platform-tests/to-upstream/css/cssom/style-set-property.html` passes.

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -96,29 +96,15 @@ const lruCache = new LRUCache({
  * Prepares a stringified value.
  *
  * @param {string|number|null|undefined} value - The value to prepare.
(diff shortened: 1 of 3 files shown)


lib/parsers.js
domenic: Replace this whole function with the following:

```js
if (value === null) {
  return "";
}
return `${value}`.trim();
```

domenic (approved):

Issue #255, opened 2025-12-05
cssstyle Refactoring Project: Split PR Plan


asamuzaK:
Planned by Gemini.
Thoughts?

# **cssstyle Refactoring Project: Split PR Plan**

**Objective:** To decouple CSSStyleDeclaration and migrate property logic to lib/properties through a series of reviewable, incremental steps.

## **Overall Roadmap**

| Order | Category | PR Title Draft | Main Changes | Impact Scope |
| :---- | :---- | :---- | :---- | :---- |
| **1** | Foundation | **Add shared parsers and property interface** | Addition of shared parsers, creation of directory structure. | No impact on existing code. |
| **2** | Core | **Refactor CSSStyleDeclaration core & registry** | Overhaul of CSSStyleDeclaration.js, implementation of the registry. | **Breaking Changes** (Unmigrated properties will temporarily stop working). |
| **3** | Migration | **Port basic properties (Color, Font, etc.)** | Migration of basic properties, fixing E2E tests. | Restoration of basic functionality. |
| **4** | Migration | **Port remaining properties (Batch 1-N)** | Mechanical/Manual migration of remaining properties. | Restoration of full functionality. |
| **5** | Cleanup | **Remove legacy generator scripts** | Removal of old scripts/generateProperties.js, etc. | Cleanup. |

### **PR 1: Foundation**

**Goal:** Prepare the common functions required by the new property definition files.

* **Changes:**  
  * Create lib/parsers/utils.js (or lib/parsers/\*.js).  
  * Implement utility functions for parsing colors (parseColor), lengths (parseLength), and processing keywords.  
  * Create the lib/properties/ directory (initially empty or with a .gitkeep).  
* **Review Points:**  
  * Are the regular expressions and logic for the parsers correct?  
  * Are unit tests (test/parsers.js, etc.) added?

### **PR 2: The Core Swap**

**Goal:** Discard the previous tightly coupled CSSStyleDeclaration and replace it with the generic container version.

* **Changes:**  
  * Rewrite lib/CSSStyleDeclaration.js to the new Map-based implementation.  
  * Create lib/properties/index.js (the registry).  
  * **Important:** Since most properties will be unregistered at this point, many existing tests may fail. Add only 1-2 sample properties (like color or display) to lib/properties/ to prove the core logic works minimally.  
* **Review Points:**  
  * Is the generic logic for setProperty / getPropertyValue correct?  
  * Is dynamic loading from the registry functioning correctly?

[...]

PR #278, opened 2026-01-02
Remove generated/properties in favor of direct mixin

asamuzaK:
## Motivation
This PR refactors the property handling mechanism.
`generated/properties.js` is a duplicate of the logic already present in `properties/*.js`.
By referencing `properties/*.js` directly from the new `generated/propertyHandlers.js`, we eliminate this redundancy and ensure a single source of truth for property definitions.

Part of https://github.com/jsdom/cssstyle/issues/255#issuecomment-3630183207
However, this PR does not include a fix for `CSSStyleDeclaration` itself.
Planning to address it in a separate PR.

## Key Changes
- **Core Logic Update**:
  - Added `scripts/generatePropertyHandlers.mjs`, which generates `generated/propertyHandlers.js` that directly references definitions in `properties/*.js`.
  - Created `./property.js` to manage property definitions dinamically.
  - Updated `CSSStyleDeclaration` to use `propertyDefinitions` and the `mixinProperties()` function from `./property.js`.
- **Cleanup**: Removed obsolete generator scripts.
- **Data Update**: Downloaded and synced the latest CSS property definitions from W3C.

### DIFF
--- a/scripts/generatePropertyDescriptors.mjs
+++ b/scripts/generatePropertyDescriptors.mjs
@@ -4,28 +4,6 @@ import propertyDefinitions from "../lib/generated/propertyDefinitions.js";
 
 const { dirname } = import.meta;
 
-// Utility to translate from `border-width` to `borderWidth`.
-// NOTE: For values prefixed with webkit, e.g. `-webkit-foo`, we need to provide
-// both `webkitFoo` and `WebkitFoo`. Here we only return `webkitFoo`.
-const dashedToCamelCase = (dashed) => {
-  if (dashed.startsWith("--")) {
-    return dashed;
-  }
-  let camel = "";
-  let nextCap = false;
-  // skip leading hyphen in vendor prefixed value, e.g. -webkit-foo
-  let i = /^-webkit-/.test(dashed) ? 1 : 0;
-  for (; i < dashed.length; i++) {
-    if (dashed[i] !== "-") {
-      camel += nextCap ? dashed[i].toUpperCase() : dashed[i];
-      nextCap = false;
-    } else {
-      nextCap = true;
-    }
-  }
-  return camel;
-};
-
 const list = fs
   .readdirSync(path.resolve(dirname, "../lib/properties"))
   .map((file) => [file.replace(/\.js$/, ""), file]);
@@ -69,3 +47,25 @@ module.exports = {
 `;
 
 fs.writeFileSync(path.resolve(dirname, "../lib/generated/propertyDescriptors.js"), output);
+
+// Utility to translate from `border-width` to `borderWidth`.
+// NOTE: For values prefixed with webkit, e.g. `-webkit-foo`, we need to provide
+// both `webkitFoo` and `WebkitFoo`. Here we only return `webkitFoo`.
+function dashedToCamelCase(dashed) {
+  if (dashed.startsWith("--")) {
+    return dashed;
+  }
+  let camel = "";
+  let nextCap = false;
+  // skip leading hyphen in vendor prefixed value, e.g. -webkit-foo
+  let i = /^-webkit-/.test(dashed) ? 1 : 0;
+  for (; i < dashed.length; i++) {
+    if (dashed[i] !== "-") {
+      camel += nextCap ? dashed[i].toUpperCase() : dashed[i];
+      nextCap = false;
+    } else {
+      nextCap = true;
+    }
+  }
+  return camel;
+}


lib/property.js
```diff
@@ -0,0 +1,62 @@
...
+    if (!Object.hasOwn(target, property)) {
+      Object.defineProperty(target, property, handler);
+    }
+  }
+  return target;
+};
+
+module.exports = {
+  mixinProperties,
+  properties: Object.fromEntries(prepareProperties())
```
domenic: I cannot find where this export is used.

lib/property.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Mixes the prepared properties into the target object.
+ * Only defines properties that are not already present on the target.
+ *
+ * @param {object} [target={}] - The target object to define properties on.
+ * @returns {object} The target object with the new properties defined.
+ */
+const mixinProperties = (target = {}) => {
+  const properties = prepareProperties();
+  for (const [property, handler] of properties.entries()) {
```
domenic: This map seems to contain pairs of `[property, descriptor]`, not `[property, handler]`.

I think we probably should not use the word "handler" anywhere. I am working through this PR to find out. This is the first place I notice the problem; I will continue commenting.

Note that in `properties/*.js`, this object is called "definition", e.g. `module.exports.definition = { set(v) { ... }, get() { ... }, enumerable: true, configurable: true }`. That is not correct either, and it would be good to fix it one day.

(See, e.g., https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor#description for reference.)

lib/property.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ * Prepares a map of properties associated with their corresponding handlers.
+ *
+ * @returns {Map<string, object>} A map where keys are property names and values are handlers.
+ */
+const prepareProperties = () => {
+  const properties = new Map();
+  for (const [key, value] of propertyDefinitions.entries()) {
```
domenic: Let's tidy this up a bit and give clearer names, like so:

```js
for (const [propertyName, { styleDeclaration }] of propertyDefinitions.entries()) {
```

asamuzaK:
PR ready.

scripts/camelize.mjs
domenic: Only dashedToCamelCase is used now. And, it is only used in generatePropertyDescriptors.mjs. Delete this file and inline that single function into generatePropertyDescriptors.mjs.

(I suggest using `function dashedToCamelCase() { ... }` instead of the `const dashedToCamelCase = () => { ... }` syntax you seem to prefer, because then you can put it at the bottom of the file, which makes the file easier to read.)

scripts/generatePropertyDescriptors.mjs
```diff
@@ -0,0 +1,54 @@
...
+  let aliasCamel = "";
+  if (legacyAliasOf) {
+    aliasCamel = dashedToCamelCase(legacyAliasOf);
+  }
+  if (camelCasedProperties.has(camel)) {
+    requires.push(`const ${camel} = require("../properties/${camel}");`);
+    for (const property of styleDeclaration) {
+      descriptors.push(`"${property}": ${camel}.definition`);
+    }
+  } else if (camelCasedProperties.has(aliasCamel)) {
```
domenic: This branch is never hit (and never should be hit; we should always ensure the files are named after their canonical property names).

Delete it and delete all the stuff around aliasCamel and legacyAliasOf.

scripts/generatePropertyDescriptors.mjs
```diff
@@ -0,0 +1,54 @@
...
+
+const outputFile = path.resolve(dirname, "../lib/generated/propertyDescriptors.js");
+
+const dateToday = new Date();
+const [dateTodayFormatted] = dateToday.toISOString().split("T");
+const output = `"use strict";
+// autogenerated - ${dateTodayFormatted}
+
+${requires.sort().join("\n")}
+const { getPropertyDescriptor } = require("../utils/propertyDescriptors");
```
domenic:
```suggestion
const { getPropertyDescriptor } = require("../utils/propertyDescriptors.js");
```

domenic (commented):
Looking great, just a few more tweaks.

domenic:
While writing up #279, I noticed something interesting we should probably fix in this PR.

Previously, `lib/generated/properties.js` contained entries for `webkitBorderAfterColor`, which was a property from `lib/utils/allExtraProperties.js`. Importantly, it is a property which we have specific handler code for, in `lib/properties/webkitBorderAfterColor.js`.

After this PR, `lib/properties/webkitBorderAfterColor.js` is not referenced anywhere, because it does not appear in the `@webref/css` database. So, it is dead code.

Per my initial thoughts in #279, I think this is OK. We can have a breaking change that removes all properties not in the `@webref/css` database.

But, we should delete the dead code ASAP. I think doing it in this PR would be best.

You should be able to quickly ask an AI to delete all files in `lib/properties/` which do not appear in `generated/propertyDescriptors.js`.

If you think deleting these files loses valuable code, then we should stop to reconsider and discuss. But I think it will probably be fine.

domenic:
> Some things have been standardized in `-webkit-foo`, so I don't think it's appropriate to simply remove everything just because it's not in `@webref/css`.

I think these two things are not in conflict. `@webref/css` contains all *standardized* `-webkit-foo` properties. For example, it contains `-webkit-flex-wrap`. It does not contain non-standardized `-webkit-foo` properties. For example, it does not contain `-webkit-border-after-color`.

asamuzaK:
> 2\. This PR should remove functionality (like it currently does), and delete all dead code.

I also prefer 2, but in a separate PR to address that issue.

scripts/generatePropertyDescriptors.mjs
asamuzaK: Removed.

scripts/generatePropertyDescriptors.mjs
```diff
@@ -0,0 +1,54 @@
...
+const { dirname } = import.meta;
+
+const list = fs
+  .readdirSync(path.resolve(dirname, "../lib/properties"))
+  .map((file) => file.replace(/\.[cm]?js$/, ""));
+const camelCasedProperties = new Set(list);
+
+const requires = [];
+const descriptors = [];
+for (const [key, value] of propertyDefinitions.entries()) {
```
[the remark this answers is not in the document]
asamuzaK: Done.

scripts/camelize.mjs
asamuzaK: > Delete this file and inline that single function into generatePropertyDescriptors.mjs.

I understand what you're saying, but I believe that a slight change in the conditions could lead to switching between dashedToCamelCase and camelCaseToDashed.

Therefore, I would like to keep camelize.js as it is, even if it contains some unnecessary function at the moment.

scripts/generatePropertyDescriptors.mjs
```diff
@@ -0,0 +1,54 @@
+import fs from "node:fs";
+import path from "node:path";
+import propertyDefinitions from "../lib/generated/propertyDefinitions.js";
+import { dashedToCamelCase } from "./camelize.mjs";
+
+const { dirname } = import.meta;
+
+const list = fs
+  .readdirSync(path.resolve(dirname, "../lib/properties"))
+  .map((file) => file.replace(/\.[cm]?js$/, ""));
```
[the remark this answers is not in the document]
asamuzaK: Removed `\.[cm]?js`, and added file extension like `properties/foo.js` in `require()`.

scripts/generatePropertyDescriptors.mjs
asamuzaK: See above.

test/CSSStyleDeclaration.test.js
```diff
@@ -27,34 +15,39 @@ describe("CSSStyleDeclaration", () => {
...
-    assert.strictEqual(typeof style.getPropertyPriority, "function");
-    assert.strictEqual(typeof style.removeProperty, "function");
+    for (const property of propertyDefinitions.keys()) {
+      const camel = dashedToCamelCase(property);
+      assert.ok(style[camel] !== undefined);
+    }
   });
 
-  it("has PascalCase for webkit prefixed properties", () => {
+  it("has PascalCase property for all webkit prefixed properties", () => {
```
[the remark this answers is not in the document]
asamuzaK: Done.

domenic:
> I'm talking about things like `-webkit-border-after-color` -> `border-block-end-color`.
> 
> It has been standardized, but renamed.

We should not include support for non-standardized aliases like `-webkit-border-after-color` -> `border-block-end-color`. Only [standardized](https://drafts.csswg.org/css-flexbox/#webkit-aliases) aliases like `-webkit-flex-wrap` -> `flex-wrap`.

> I also prefer 2, but in a separate PR to address that issue.

OK. So will you do 1 in this PR?

scripts/generatePropertyDescriptors.mjs
asamuzaK: This process is necessary.
After fixing #279, it will look like this.
```
{
  "-weblkit-flex": flex.definition,
  "webkitFlex": flex.definition,
  "WebkitFlex": flex.definition
}
```

domenic (commented):
Almost perfect; thanks so much for your work here. The webkit property handling looks solid.

scripts/camelize.mjs
asamuzaK: Since there's no point in arguing over a matter of code styling preference, I've decided to implement it as you suggested.

However, I think it would be better to define it in eslint [func-style](https://eslint.org/docs/latest/rules/func-style).
Furthermore, it would be better to unify the code style with jsdom?

domenic (approved):

PR #271, opened 2025-12-27
Add LRU cache for CSS parsing and validation

asamuzaK:
Summary
Adds an LRU cache to reduce repeated css-tree parsing/validation and improve performance of CSS parsing utilities.

Part of https://github.com/jsdom/jsdom/issues/3985

What this PR does
- Adds lib/utils/cache.js: simple LRU wrapper using lru-cache (max: 4096) with a null sentinel for null values.
- Caches results in lib/parsers.js for:
  - isValidPropertyValue (key: isValidPropertyValue_${prop}_${val}) — caches boolean results
  - resolveCalc (key: resolveCalc_${val}) — caches resolved string value
  - parsePropertyValue (key: parsePropertyValue_${prop}\_${val}\_${caseSensitive}) — caches parsed AST/plain-object results or false when invalid
- Adds lru-cache to package.json and package-lock.json.

Benchmarks
cssstyle main:
```
> jsdom@27.4.0 benchmark
> node ./benchmark/runner --suites style

# style/createElement + setAttribute('style') #
jsdom  x 44.33 ops/sec ±1.27% (59 runs sampled)

# style/createElement + style properties #
jsdom  x 142 ops/sec ±3.34% (72 runs sampled)

# style/createElement + style.cssText #
jsdom  x 36.25 ops/sec ±2.77% (59 runs sampled)

# style/innerHTML: divs with inline styles #
jsdom  x 41.36 ops/sec ±2.46% (55 runs sampled)

# style/innerHTML: simple divs (no styles) #
jsdom  x 3,739 ops/sec ±1.71% (88 runs sampled)
```

cssstyle PR:
```
> jsdom@27.4.0 benchmark
> node ./benchmark/runner --suites style

# style/createElement + setAttribute('style') #
jsdom  x 178 ops/sec ±1.46% (83 runs sampled)

# style/createElement + style properties #
jsdom  x 248 ops/sec ±3.14% (78 runs sampled)

# style/createElement + style.cssText #
jsdom  x 148 ops/sec ±3.85% (76 runs sampled)

# style/innerHTML: divs with inline styles #
jsdom  x 154 ops/sec ±2.61% (79 runs sampled)

# style/innerHTML: simple divs (no styles) #
jsdom  x 3,837 ops/sec ±2.53% (89 runs sampled)
```

PR #284, opened 2026-01-06
Refactor CSSStyleDeclaration

asamuzaK:
## **Motivation**

This PR refactors CSSStyleDeclaration to better match the internal architecture requirements for jsdom integration.
Key changes include overhauling the constructor, removing the setter for length, and converting internal methods to class methods.

Part of https://github.com/jsdom/cssstyle/issues/255#issuecomment-3630183207

### **1\. Constructor & Initialization Updates**

* **Direct Property Assignment**: Replaced Object.defineProperties with direct property assignments for initializing internal state, improving readability and performance.
* **Internal Property Cleanup**:  
  * Renamed \_setInProgress to \_updating for clarity.  
  * Grouped internal properties with comments (e.g., "Internals for jsdom", "Internals for CSS declaration block").

### **2\. Length Property**

* **Remove set length**: Removed the setter for length. The property is now read-only, consistent with the CSSOM interface.
* **Replace splice**: Array.prototype.splice calls were removed because they attempt to set the read-only length property.
  * Implemented \_clearIndexedProperties() to handle full clears.  
  * Implemented \_removeIndexedProperty(index) to handle single item removal and index shifting.
  * Although it is not directly related to length setter, Array.prototype.splice.call(this, property) are also removed and implemented as \_getIndexOf(property).

### **3\. Internal Methods Refactoring**

* **Class Methods**: Moved internal methods (like \_setProperty, \_borderSetter, \_flexBoxSetter, etc.) from Object.defineProperties(CSSStyleDeclaration.prototype, ...) directly into the class definition.
* **Argument Renaming**: Renamed the prior argument to priority in all setters for consistency.
  * Also renamed prop to property and val to value.

### **4\. Documentation**

* **JSDoc Overhaul**: Updated JSDoc comments throughout the file to provide clearer API documentation.

### DIFF
--- a/lib/CSSStyleDeclaration.js
+++ b/lib/CSSStyleDeclaration.js
@@ -33,9 +33,7 @@ class CSSStyleDeclaration {
    * @param {object} [opt] - Options.
    * @param {object} [opt.context] - The context object (Window, Element, or CSSRule).
    */
-  constructor(onChangeCallback, opt = {}) {
-    const { context } = opt;
-
+  constructor(onChangeCallback, { context } = {}) {
     // Internals for jsdom
     this._global = globalThis;
     this._onChange = onChangeCallback;


asamuzaK:
https://github.com/jsdom/cssstyle/blob/87e5577ef9535b453043d34460c11be626c562df/lib/CSSStyleDeclaration.js#L131
This line runs the parser every time set cssText() is called, so caching it here should result in a slight performance improvement.

However, this contradicts what was previously stated in https://github.com/jsdom/cssstyle/pull/271#issuecomment-3701835473. What do you think?

lib/CSSStyleDeclaration.js
```diff
@@ -27,83 +27,32 @@ const { asciiLowercase } = require("./utils/strings");
...
-        enumerable: false,
-        writable: true
-      },
-
-      _priorities: {
-        value: new Map(),
-        enumerable: false,
-        writable: true
-      },
+    const { context } = opt;
```
domenic: Shorten these lines by doing `constructor(onChangeCallback, { context }) {`

lib/CSSStyleDeclaration.js
```diff
@@ -27,83 +27,32 @@ const { asciiLowercase } = require("./utils/strings");
...
 
-      _length: {
-        value: 0,
-        enumerable: false,
-        writable: true
-      },
+    // Internals for jsdom
+    this._global = globalThis;
+    this._onChange = typeof onChangeCallback === "function" ? onChangeCallback : null;
```
domenic: Let's simplify
```suggestion
    this._onChange = onChangeCallback;
```

If the caller wants to pass null (or undefined) they can do so directly, instead of allowing any non-function value to trigger this.

lib/CSSStyleDeclaration.js
```diff
@@ -157,22 +108,28 @@ class CSSStyleDeclaration {
...
+   * @param {string} text - The new CSS text.
+   */
+  set cssText(text) {
     if (this._readonly) {
       const msg = "cssText can not be modified.";
       const name = "NoModificationAllowedError";
       throw new this._global.DOMException(msg, name);
     }
-    Array.prototype.splice.call(this, 0, this._length);
+    this._clearIndexedProperties();
```
domenic: I really like these improvements to indexed property handling. Centralizing that all makes it much more clear.

lib/CSSStyleDeclaration.js
```diff
@@ -335,29 +309,32 @@ class CSSStyleDeclaration {
...
       const name = "NoModificationAllowedError";
       throw new this._global.DOMException(msg, name);
     }
-    const value = prepareValue(val);
+    value = prepareValue(value);
     if (value === "") {
-      this[prop] = "";
-      this.removeProperty(prop);
+      if (Object.hasOwn(propertyDescriptors, property)) {
+        propertyDescriptors[property].set.call(this, value);
```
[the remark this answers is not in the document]
asamuzaK: Done.

lib/CSSStyleDeclaration.js
```diff
@@ -366,260 +343,285 @@ class CSSStyleDeclaration {
...
+      this[this._length] = property;
+      this._length++;
+    }
+    if (priority === "important") {
+      this._priorities.set(property, priority);
+    } else {
+      this._priorities.delete(property);
+    }
+    this._values.set(property, value);
+    if (typeof this._onChange === "function" && !this._updating && this.cssText !== originalText) {
```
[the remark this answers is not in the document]
asamuzaK: Done, also elsewhere.

lib/CSSStyleDeclaration.js
domenic: Sorry, I meant to type `constructor(onChangeCallback, { context } = {})`

domenic (approved):
To avoid extra delay I will integrate the parameter change myself and merge

PR #289, opened 2026-01-09
perf: Fix parsePropertyValue cache to include inArray parameter

domenic:
The cache key was not including the inArray parameter, which meant that cache entries would be incorrectly reused between calls with different inArray values, causing unnecessary re-parsing.

Since almost all callers use inArray: true, adding it to the cache key significantly improves cache hit rates and overall performance.

---

To-do before merging:
- [ ] Human code review (@asamuzaK welcome to help)
- [ ] Double-check benchmark improvements manually
- [ ] Double-check the full jsdom test suite passes

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -265,15 +265,11 @@ const parsePropertyValue = (prop, val, opt = {}) => {
     }
     val = calculatedValue;
   }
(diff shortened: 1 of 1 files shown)


domenic:
See https://github.com/jsdom/cssstyle/pull/290

Issue #280, opened 2026-01-03
Cleanups after removing generated/properties


domenic:
Followups after #278:

- [x] Rename each handler's `description` export to `descriptor` or `propertyDescriptor`, for clarity.
- [ ] Rename `getPropertyDescriptor()` to `createGenericPropertyDescriptor()` for clarity, and add documentation explaining how it is used when we don't have handler code.
- [x] Reduce the size of the package (as seen by `npm pack`). Either implement #279, or:
  - [x] Remove `lib/utils/allExtraProperties.js`. It is only used by prepare scripts.
  - [x] Remove `lib/generated/allProperties.js`. It is only used by prepare scripts.

asamuzaK:
Add this to the list too: 
* Remove unnecessary `globalObject` argument passing to parsers functions, e.g. `parsers.parsePropertyValue(value, { globalObject: this._global })`.
  Need to fix:
  * `lib/properties/*.js`
  * `lib/utils/propertyDescriptor.js`

PR #290, opened 2026-01-09
perf: Always return arrays from parsePropertyValue

domenic:
The inArray parameter caused cache thrashing: CSSStyleDeclaration.js
called without it (caching strings) while property files called with
inArray: true (expecting arrays). When the cached type didn't match
what a caller needed, it would recompute and overwrite, causing the
next different-type caller to also miss. By always returning arrays,
all callers can share the same cached results.

---

To-do before merging:
- [x] Human code review (@asamuzaK welcome to help)
- [x] Double-check benchmark improvements manually
- [x] Double-check the full jsdom test suite passes

### DIFF
--- a/lib/properties/backgroundAttachment.js
+++ b/lib/properties/backgroundAttachment.js
@@ -14,8 +14,7 @@ module.exports.parse = (v, opt = {}) => {
   const parsedValues = [];
   for (const val of values) {
     const value = parsers.parsePropertyValue(property, val, {
-      globalObject,
-      inArray: true
+      globalObject
     });
     if (Array.isArray(value) && value.length === 1) {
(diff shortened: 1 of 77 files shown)


lib/properties/backgroundAttachment.js
```diff
@@ -14,8 +14,7 @@ module.exports.parse = (v, opt = {}) => {
   const parsedValues = [];
   for (const val of values) {
     const value = parsers.parsePropertyValue(property, val, {
-      globalObject,
-      inArray: true
+      globalObject
```
asamuzaK: We can remove all `{ globalObject }`, not just this file.
See https://github.com/jsdom/cssstyle/issues/280#issuecomment-3707846197

PR #295, opened 2026-01-10
perf: Skip redundant border shorthand expansion during cssText updates

domenic:
When cssText is set, prepareProperties() already expands border shorthands into all their longhand properties. Previously, calling setProperty() on these shorthands would then trigger their property setters, which called _borderSetter() to expand them again - redundant work.

This makes border property setters check the _updating flag (set during cssText operations) and skip _borderSetter() when true. The setters still call parse() to validate and canonicalize the value, then store it directly via _setProperty().

Affected: border, border-width, border-style, border-color, border-top, border-right, border-bottom, border-left

---

```
  Parent commit (without optimization):
  - style/createElement + style.cssText: 221 ops/sec
  - style/innerHTML: divs with inline styles: 198 ops/sec

  With optimization:
  - style/createElement + style.cssText: 360 ops/sec (+63%)
  - style/innerHTML: divs with inline styles: 305 ops/sec (+54%)
```

---

@asamuzaK I am unsure on this one. It adds a good amount of repetitive code, and is subtle and hard to understand. I guess we can probably factor a bit out to reduce the repetition (I might push some commits to try to do that). But I'm still unsure.

The main thing I'm curious about is how this fits with your larger plans in https://github.com/jsdom/cssstyle/issues/255. Do you plan to overhaul these handlers anyway, onto a newer/better architecture? Should we just wait for that? Or would this sort of optimization be valuable longer term?

I also tried to see if this generalizes to other properties besides the border ones, but am not having too much luck so far.

### DIFF
--- a/lib/properties/border.js
+++ b/lib/properties/border.js
@@ -92,7 +92,15 @@ module.exports.definition = {
       });
       if (val || typeof val === "string") {
         const priority = this._priorities.get(property) ?? "";
-        this._borderSetter(property, val, priority);
+        if (this._updating) {
+          const shorthandValue =
+            typeof val === "string"
+              ? val
+              : (Array.isArray(val) ? val : Object.values(val)).join(" ");
+          this._setProperty(property, shorthandValue, priority);
+        } else {
+          this._borderSetter(property, val, priority);
+        }
       }
     }
   },
--- a/lib/properties/borderBottom.js
+++ b/lib/properties/borderBottom.js
@@ -85,7 +85,15 @@ module.exports.definition = {
           !this._priorities.get(shorthand) && this._priorities.has(property)
             ? this._priorities.get(property)
             : "";
-        this._borderSetter(property, val, priority);
+        if (this._updating) {
+          const shorthandValue =
+            typeof val === "string"
+              ? val
+              : (Array.isArray(val) ? val : Object.values(val)).join(" ");
+          this._setProperty(property, shorthandValue, priority);
+        } else {
+          this._borderSetter(property, val, priority);
+        }
       }
     }
   },
--- a/lib/properties/borderColor.js
+++ b/lib/properties/borderColor.js
@@ -95,7 +95,15 @@ module.exports.definition = {
           !this._priorities.get(shorthand) && this._priorities.has(property)
             ? this._priorities.get(property)
             : "";
-        this._borderSetter(property, val, priority);
+        if (this._updating) {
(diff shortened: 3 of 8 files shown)


domenic:
> CSS borders have two shorthands: line (border-color, border-style, border-width) and position (border-top, border-right, border-bottom, border-left). This makes the process redundant.
> 
> At this point, I have no idea how to improve on what is already implemented in normalize.js and parsers.js.

Can you explain a bit more, since you understand this project better than me? In particular, how normalize.js and parsers.js map to the handler files?

For example... Do we expect eventually every CSS property to have its own handler file? Or will we be able to use some generic patterns, like a line handler and a position handler? Or will we maybe generate handlers from the grammars in `@webref/css`?

asamuzaK:
There are at least two shorthands that require special handling in CSS.

One is the background, which allows multiple backgrounds.
The other is the border.
While the regular shorthands are simply shorthands for the longhands and have a one-to-one correspondence, e.g. margin is a shorthand for margin-top, margin-right, margin-bottom and margin-left, border has intermediate shorthands in between, and there are two line and position shorthands.

To make things even more confusing, border specifies all four sides with a single value, but border-style and border-top etc. allow you to specify any of sides 1 to 4 individually.

So, If we have the following for example:
```
border: 1px solid red; border-style: dotted; border-top: 2px double; border-bottom-color: green;
```

We need to follow these steps in the order the CSS is specified:
1. First, expand border shorthand to all longhands
2. Then, override the corresponding longhands with the border-style value
3. Then, override the corresponding longhands with the border-top value
4. Then, override border-bottom-color
5. Finally, if there are any shorthand values that can be specified, set them to those values.

asamuzaK:
> Although, border-block-end-color is "\<color\> | \<image-1D\>"... I guess the latter works reasonably well now because of the fallback here?

Hmm, not quite.
`<image-1D>` is actually the CSS function `stripes()`.
Therefore, it matches the `if (Array.isArray(value) && value.length === 1) {` block and is passed to `parsers.resolveColorValue(value)`.
As a result, it is treated as an invalid value and undefined is returned.

Currently, no browsers support stripes(), and there is no test for it in wpt.

While `<image-1D>` is not a "color", it stores color data internally, similar to gradients, so I'm considering supporting it in cssColor.

domenic:
I'll give up on this and maybe try again one day. Now that the tests are in the same repo it should be easier.

PR #298, opened 2026-01-11
Remove workaround for csstree

asamuzaK:
https://github.com/w3c/webref/issues/1647 is fixed.


asamuzaK:
css-syntax-patches-for-csstree had a bug.
https://github.com/csstools/postcss-plugins/issues/1760

When releasing a new cssstyle, make sure to update it again.

domenic:
Can you send an update PR with a test that would catch the bug? (Maybe such a test already exists and that's how you caught it.)

domenic:
Sure, that sounds good. Thanks for the detailed explanation.

PR #300, opened 2026-01-12
Align code style with jsdom

### LINKED ISSUE #286: Unify code style with jsdom
Per https://github.com/jsdom/cssstyle/pull/278#discussion_r2663512564

---
Thanks for following my requests here. I'm happy to discuss this in more detail in a friendly manner :).

Personally, I think linear flow for code is an anti-goal. It is more important for code reading that the code central to a file be at the top and that other code be at the bottom, in my opinion.

I do agree consistency is important. So I think defining it in ESLint one way or another is a good idea.

For cssstyle, since you are doing most of the work, I am happy to let you set the style guidelines. That is why I put the part about function style as "I suggest"; the part I felt strongly about was no dead code, but the style was just a suggestion. I am sorry if that was not clear.

If you want to unify with jsdom, I think that is a great idea. (jsdom [uses `func-style` to enforce declarations](https://github.com/domenic/eslint-config/blob/76faba551ae839009c1748500829008892d17546/index.mjs#L84).) Of course, the jsdom codebase is all using my personal style. But if you would prefer to set your own style for this project, that is fine too.

_Originally posted by @domenic in https://github.com/jsdom/cssstyle/pull/278#discussion_r2663512564_

asamuzaK:
Fix #286

## Motivation

We will apply a code style that largely matches jsdom's.
The objective is to unify code styling across the jsdom project to improve readability and maintainability, as well as to facilitate debugging. 

## Changes from the existing code style

* Copied and applied `.editorconfig` from jsdom.
* Aligned rules in `eslint.config.mjs` to nearly match jsdom's.
* Use function declarations (`function foo() {}`) instead of arrow function expressions (`const foo = () => {}`).
* Unified the file structure to the following order:
  * Imports (`require()`)
  * Top-level Constants & Configuration
  * Functions
  * Module Exports

## Differences from jsdom

* We will continue using Prettier.
  * Added `"consistent-return": "off"` to the `rules` to accommodate Prettier usage.
  * As a side note, I personally believe jsdom should consider using Prettier as well.
  * `"func-style": "off"` for test files.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -9,7 +9,8 @@ export default [
   {
     files: ["**/*.js"],
     languageOptions: {
-      sourceType: "commonjs"
+      sourceType: "commonjs",
+      globals: globals.node
     }
   },
   ...domenicConfig,
@@ -18,21 +19,6 @@ export default [
     rules: {
       "consistent-return": "off",
       "require-unicode-regexp": "off"
-    }
-  },
-  {
-    files: ["scripts/**/*"],
-    languageOptions: {
-      globals: globals.node
-    }
-  },
-  {
-    files: ["test/**/*"],
-    languageOptions: {
-      globals: globals.node
-    },
-    rules: {
-      "func-style": "off"
     }
   }
 ];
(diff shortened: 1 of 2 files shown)


eslint.config.mjs
```diff
@@ -35,6 +38,9 @@ export default [
     files: ["test/**/*"],
     languageOptions: {
       globals: globals.node
+    },
+    rules: {
+      "func-style": "off"
```
domenic: Why not also update test files? What is the benefit of the inconsistency?

domenic (approved):

PR #321, opened 2026-09-10
Reject negative lengths for border-spacing

afonsojanu:
## Summary

`border-spacing` currently accepts and stores negative lengths, e.g.:

```js
style.borderSpacing = '-10px';
style.borderSpacing; // '-10px'
```

That's not spec-compliant. The CSS Tables spec defines the property as `<length [0,∞]>{1,2}` and its prose says the values "must be non-negative" (this carries over from CSS 2.1, which put it even more plainly: "Negative values for `<length>` are illegal").

Looking through the other length properties in this codebase that carry the same non-negative constraint, they all pass `min: 0` into `resolveNumericValue`/`serializeLength`: `border-*-width`, `padding*`, `width`, `height`, `line-height`, `flex-grow`/`flex-shrink`. `border-spacing` is the one property with that same constraint that was missing it, in both the single-value and two-value branches of its parser.

## Fix

Adds `min: 0` to the `resolveNumericValue` calls in `lib/properties/borderSpacing.js`, for both the one-length and two-length forms. With this in place, a negative length now falls through to `undefined` the same way it already does for the sibling properties, and the whole property is left unset (matching how e.g. `border-top-width: -1px` already behaves here).

## Test plan

- Added three cases to `test/properties.test.js`: a single negative length, a negative length as the first of two values, and as the second of two values. All assert the property does not get set.
- Verified these three tests fail on the old code (`'-10px' !== ''`, etc.) and pass with the fix.
- `node --test` passes (664/664).
- `eslint` is clean on both changed files.

## Disclosure

I used an AI coding assistant (Claude) to help investigate and write this fix, per this project's contribution norms.


domenic:
This project is no longer maintained; it has been rolled into jsdom. Good reminder to archive it though.

